PR #6, opened 2019-01-03
 MIVisionX Setup Script Fix

kiritigowda:
* sudo requirement removed
* fixed bugs

### DIFF
--- a/MIVisionX-setup.py
+++ b/MIVisionX-setup.py
@@ -1,50 +1,32 @@
 __author__      = "Kiriti Nagesh Gowda"
 __copyright__   = "Copyright 2018, AMD Radeon MIVisionX setup"
 __license__     = "MIT"
-__version__     = "0.9.9"
+__version__     = "0.9.91"
 __maintainer__  = "Kiriti Nagesh Gowda"
 __email__       = "Kiriti.NageshGowda@amd.com"
 __status__      = "beta"
 
+import argparse
 import os
-import getopt
 import sys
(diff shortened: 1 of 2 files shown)


jlgreathouse:
LGTM for the the issues it's aiming to fix, thanks!

What are your thoughts of putting `#! /usr/bin/env python` on the first line and `chmod +x` the script so that it could be run directly (`./MIVisionX-setup.py`) instead of having to call `python MIVisionX-setup.py`?

kiritigowda:
> LGTM for the the issues it's aiming to fix, thanks!
> 
> What are your thoughts of putting `#! /usr/bin/env python` on the first line and `chmod +x` the script so that it could be run directly (`./MIVisionX-setup.py`) instead of having to call `python MIVisionX-setup.py`?

@jlgreathouse Thanks for reporting these issues. I will merge this to the master. Please do report any problems or bugs you encounter. 

Any particular reason for having `./MIVisionX-setup.py` instead of the `python MIVisionX-setup.py`?

jlgreathouse:
@kiritigowda on Linux anyway, it's usually common practice for scripts to be executable directly. It's the whole reason for [the hashbang mechanism](https://en.wikipedia.org/wiki/Shebang_(Unix)) is to treat scripts as executables so you don't need to personally know what interpreter is being used.

kiritigowda:
@jlgreathouse thanks for the information, I will make those changes in my next commit.

PR #31, opened 2019-01-23
OpenVX WinML Extension

kiritigowda:
* This merge will include OpenVX WinML extension, which will allow developers to run pre-trained ONNX models to be loaded directly to the OpenVX graph and run the neural net model with pre & post processing vision fuctions from OpenVX and OpenCV extensions.


kiritigowda:
@LakshmiKumar23 this is functionally complete. Please run the samples to test.

PR #83, opened 2019-04-08
relative path decode for runvx

paveltc:
added feature for runvx to work from the cwd of the gdf rather than the cwd of where runvx is being called from.


LakshmiKumar23:
@paveltc Can you change the [README](https://github.com/GPUOpen-ProfessionalCompute-Libraries/MIVisionX/tree/master/utilities/runvx#canny-edge-detector)  here to the same as how the gdf should look now?

PR #101, opened 2019-04-19
Support for ONNX V1.3

LakshmiKumar23:


LakshmiKumar23:
Tested using resnet50 ONNX Model V1.3

kiritigowda:
@LakshmiKumar23 does all the 1.3 Models run?

LakshmiKumar23:
@kiritigowda No. Only the resnet50 runs. The others have some issues with some layers. I will look into it

PR #102, opened 2019-04-27
add back the print needed for model_compiler output

rrawther:
This is used in inference app.


kiritigowda:
@LakshmiKumar23 could you run ONNX and Caffe profiling on this?

kiritigowda:
PR #103

PR #134, opened 2019-06-03
fixed gemm input initialization in nnir.py

paveltc:
Added the line:
input = self.tensor_dict[node.inputs[0]]
to the GEMM node branch.

### DIFF
--- a/model_compiler/python/nnir.py
+++ b/model_compiler/python/nnir.py
@@ -275,22 +275,20 @@ class IrGraph:
         for node in self.nodes:
             for output in node.outputs:
                 count+=1
+                input = self.tensor_dict[node.inputs[0]]
                 if node.type in ['sum', 'add', 'sub', 'mul', 'muladd', 'batch_norm', 'relu', 'leaky_relu', 'softmax']:
-                    input = self.tensor_dict[node.inputs[0]]
                     local = IrTensor()
                     local.setName(output)
                     local.setInfo(input.type, input.shape)
                     local.setFormat(input.format)
                     self.addLocal(local)
                 elif node.type in ['global_avg_pool']:
-                    input = self.tensor_dict[node.inputs[0]]
                     local = IrTensor()
                     local.setName(output)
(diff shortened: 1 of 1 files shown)


model_compiler/python/nnir.py
```diff
@@ -331,6 +331,7 @@ def updateLocals(self):
                     local.setFormat(input.format)
                     self.addLocal(local)
                 elif node.type in ['gemm']:
+                    input = self.tensor_dict[node.inputs[0]]
```
hansely: Variable A is also assigned as self.tensor_dict[node.inputs[0]] at line 335.
Can we just use A instead?

hansely: line 351: local.setInfo(input.type, output_shape) -> local.setInfo(A.type, output_shape)
line 352: local.setFormat(input.format) -> local.setFormat(A.format)

paveltc: @hansely The problem that I am seeing occurs because input is uninitialized for this node. I suppose A can be used instead, but more changes would be needed.

hansely: The variable 'input' is only used twice in this case(gemm operation) which is line 351 and 352.

hansely: If we want consistency over all the layers, we can use 'input' instead of A .
So it would be

input = self.tensor_dict[node.inputs[0]]
B = self.tensor_dict[node.inputs[1]]
shapeA = input.shape
.
.
.
local.setInfo(input.type, output_shape)
local.setFormat(input.format)

paveltc: @rrawther Can you please comment on this? I think @hansely has a valid concern.

rrawther: Should we just take out input = self.tensor_dict[node.inputs[0]] outside the for loop like this, that is more cleaner
        for node in self.nodes:
            input = self.tensor_dict[node.inputs[0]]
Also remove the statement from every if condition

paveltc: I changed it so that the input initialization is outside of the if statements. It should be good to merge now.

PR #142, opened 2019-06-18
fix bug in CBA fusion and added fusion for leaky relu

rrawther:
fix bug in convolution_layer for calling fusion api
added fusion mode support in nnir for leaky_relu


kiritigowda:
@paveltc is this branch ready to be merged?

PR #146, opened 2019-07-02
Visual Profiler for VX_NN

kiritigowda:
* Visual Profiler added to AMD OpenVX NN
* Profiler turned off by default
* Turn ON Profiler
   * profiler.h - PROFILER_MODE=1


kiritigowda:
@LakshmiKumar23 & @hansely can you try this branch and let me know if this works for you?

PR #150, opened 2019-07-15
printing out local buffers from anntest.py

LakshmiKumar23:
Part 2 of calibration


LakshmiKumar23:
Closing. Removing unnecessary Local/Output print functions.

PR #167, opened 2019-09-06
Fixed FindTurboJpeg cmake and set rpp installation default to no

Reza-Najafi:


MIVisionX-setup.py
```diff
@@ -16,7 +16,7 @@
 parser.add_argument('--installer', type=str, default='apt-get', help='Linux system installer - optional (default:apt-get) [options: Ubuntu - apt-get; CentOS - yum]')
 parser.add_argument('--miopen',    type=str, default='2.0.0',   help='MIOpen Version - optional (default:2.0.0)')
 parser.add_argument('--ffmpeg',    type=str, default='no',      help='FFMPEG Installation - optional (default:no) [options: Install ffmpeg - yes')
-parser.add_argument('--rpp',       type=str, default='yes',     help='Radeon Performance Primitives (RPP) Installation - optional (default:yes) [options: Install rpp - yes]')
+parser.add_argument('--rpp',       type=str, default='no',     help='Radeon Performance Primitives (RPP) Installation - optional (default:yes) [options: Install rpp - yes]')
```
kiritigowda: This will not install RALI by default, any reason to turn it off?

PR #188, opened 2019-10-01
ADAT fixes

LakshmiKumar23:


model_compiler/python/nnir.py
```diff
@@ -1164,7 +1164,7 @@ def toFile(self,outputFolder):
             os.mkdir(outputFolder)
         irDescFile = outputFolder + '/old_graph.nnir'
         irDescFileNew = outputFolder + '/graph.nnir'
-        print('OK: creating IR description in ' + irDescFile + ' ...')
+        print('OK: creating IR description in ' + irDescFileNew + ' ...')
```
kiritigowda: did you want to check in these files?

LakshmiKumar23: Yes! Just changes the print from old_graph.nnir to graph.nnir

rrawther: The code looks hacky to me. Why we need 2 separate files? I think tofile() is writing to temp_file and reading it and fixing something. Please revisit the code and fix it.

PR #194, opened 2019-10-11
Rr/multi thread linux

rrawther:
Add graph level multi-thread support for non windows platform.


kiritigowda:
@paveltc can we run tests on this PR in windows & Linux?

rrawther:
@paveltc  Fixed bug reported. Please retest.

paveltc:
@rrawther It works fine now.

PR #199, opened 2019-10-23
Added loop parameter to RALI to go through the images in a loop.

Reza-Najafi:


kiritigowda:
@Reza-Najafi can you fix the merge conflicts.

PR #211, opened 2019-11-14
yolov3 model compiler fix

rrawther:
fix for yolov3 upsample layer from caffe model


kiritigowda:
@paveltc can you run some tests on this branch and let me know if I can merge this?

Issue #172, opened 2019-09-09
MIVisionX-setup.py script fails on Python 3.X


LakshmiKumar23:
'commands' module has deprecated in Python 3.X. Replaced by 'subprocess' module. Script fails.

kiritigowda:
#216 needs to be verified to close this bug

PR #216, opened 2019-11-21
making setup script python 3 compatible

LakshmiKumar23:
Resolves issue #172


kiritigowda:
@hansely can you run this on your system and verify for python3?

PR #222, opened 2019-12-11
added command line batch size to the rali performance test app and ad…

paveltc:
…ded a modified app for rali performance that takes both batch size and depth size as input


kiritigowda:
@paveltc can you sync with TOT master before merge.

Issue #223, opened 2019-12-12
Optimize Package Size


kiritigowda:
* Remove duplicates
* Consolidate images and videos
* Remove images larger than 1 MB
* Readme images are duplicated across folders, need to get them consolidated into doc/images

kiritigowda:
level-1 optimizations merged.

PR #226, opened 2019-12-13
optimizing data consolidation

LakshmiKumar23:
- Removed duplicate images from README and moved them to docs/images
- Level 2 clean up for data consolidation
-  Fixing issue #223


LakshmiKumar23:
@kiritigowda Still need to check image sizes and reduce them. Will do it as a new phase

PR #229, opened 2019-12-19
resizing images greater than 1MB in READMEs

LakshmiKumar23:


LakshmiKumar23:
Fixes issue #223. Phase 3 (last phase) complete

PR #230, opened 2019-12-19
lrn layer

hansely:


hansely:
MIVisionX build passed

kiritigowda:
Will close this to keep, but keep this branch. We have to update VX headers before we can merge.

PR #237, opened 2020-01-20
updating cast layer to support ssd-mobilenet

LakshmiKumar23:
Adds support for attribute to = 1 (float32 to float32 casting)


hansely:
Checked with Lakshmi yesterday and looks good to me

PR #240, opened 2020-01-25
H/constant layer

hansely:
-Implemented constant layer for onnx support
-There are some nodes in onnx that doesn't have input(e.g. constant). Therefore, moved input = self.tensor_dict[node.inputs[0]] to every nodes with input.


kiritigowda:
@hansely can I merge this?

PR #250, opened 2020-03-12
porting to OpenVX_1.3

LakshmiKumar23:
These have changes from Hansel and me.


LakshmiKumar23:
Sent PR to wrong thread

PR #257, opened 2020-03-30
Enabling batch processing support , Adding new Python API

Reza-Najafi:


kiritigowda:
@Reza-Najafi every file in MIVisionX is changed in this PR. Can you check-in files with changes only.

Issue #265, opened 2020-04-28
agoReleaseGraph hangs with multithreading


hansely:
agoReleaseGraph function in openvx/ago/ago_interface.cpp hangs when used with latest multithreading code

rrawther:
Fixed with resent PR

kiritigowda:
#266

PR #266, opened 2020-04-29
fix for issue #265

rrawther:
Fixes the hang for issue #265


kiritigowda:
@hansely can you check if this fix works?

hansely:
It still hangs sometimes. It's the timing issue. 
If the wait() function is called before notify_one() is called, then it works. However, if notify_one() is called before wait(), then it hangs.

hansely:
So basically, if the thread working on releaseGraph() function starts working even before the createGraph() is finished, then it hangs.

hansely:
Nevermind. I was running a different branch. Please disregard my previous comments. 
This fixes the bug.

PR #274, opened 2020-05-05
Setup Script - RPP turn off on CentOS

kiritigowda:


kiritigowda:
Required for #274

LakshmiKumar23:
@kiritigowda Builds successfully on centos(without rpp) and ubuntu(with rpp)

PR #276, opened 2020-05-07
fixed bug in rali_unittests.cpp

paveltc:


paveltc:
This PR simply fixes a bug in the command line interface for the rali_unittests app. It does not affect any other code.

PR #289, opened 2020-05-20
moved all rali utilities into rali folder, added rali-python-unittest.py

paveltc:


paveltc:
@kiritigowda Can you please merge this PR? It just organizes all of the rali utilities under a rali directory. There's also a script added but it doesn't affect any of the code.

utilities/rali/rali_performance_tests_with_depth/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
+################################################################################
```
kiritigowda, line 1: @paveltc the changes look good. I was wondering if we can change it to rali/performance_tests... so on, having rali/rali_ on all folders would be redundant. May be we can rename the folders without rali_.Let me know what you think.

kiritigowda:
@paveltc can we look into [LGTM warnings](https://lgtm.com/projects/g/GPUOpen-ProfessionalCompute-Libraries/MIVisionX/rev/pr-0aea023b24fa4e1da12be671feffbfe7fa614a85).

paveltc:
@kiritigowda I checked the LGTM warnings and they all have to do with unused imports and a couple of unused variables in the script. I'd rather keep that as is because I am using a different script as a starting point and I don't want to remove code that might be needed soon.
As far as renaming the tests in the rali subfolder, I'd rather not do that simply because the rali prefix helps make clear that it's a rali related test. We have other unit tests and performance tests and it just makes it easier to read if the rali prefix is there.

PR #291, opened 2020-05-27
fix for random hang when shuffle is set

rrawther:


paveltc:
@kiritigowda This should be fine to merge now.

PR #294, opened 2020-05-28
RALI - Ubuntu 20.04 Fix

kiritigowda:


kiritigowda:
#292 - fixes the RALI Error

PR #317, opened 2020-06-11
CMakeList Fix - OSX Support

kiritigowda:
* Find make fixes
* OSX Support
* Warnings fix


kiritigowda:
@LakshmiKumar23 does this do ok on Linux?

PR #323, opened 2020-06-16
nnir bug fix

LakshmiKumar23:
reading from graph.nnir: Reads a list to find int/float values. Changed from looking for numerical digits to count(.) in floats.


paveltc:
@LakshmiKumar23 This passed the smoke tests that I ran. Should be fine to merge.

LakshmiKumar23:
@kiritigowda Pavel ran the smoke tests. Can you merge this?

PR #324, opened 2020-06-22
Rr/rali shard padding

rrawther:
Fixes hang in multi-gpu training


kiritigowda:
@paveltc can we merge this?

PR #342, opened 2020-07-23
gather fix

hansely:
-slight fix in the type of 'axis' parameter


hansely:
Passed the gather layer test that I previously used

PR #393, opened 2020-11-13
Loom Shell Multi-GPU support

kiritigowda:
* Loom Shell works fine on a single GPU system
* Loom Shell on the multi-GPU system was not enabled
* This fix supports multi-GPU but uses GPU[0] to run the workload


paveltc:
@kiritigowda I ran the tests. This fixes the bug and it is ready to merge.

PR #403, opened 2020-12-08
Fix mivisionx build warning

asalmanp:
* Define CL_TARGET_OPENCL_VERSION to 220 to fix warnings when building the MIVisionX toolkit.


rrawther (approved):
please make sure tests pass before merging

amd_openvx/openvx/CMakeLists.txt
```diff
@@ -72,6 +72,7 @@ add_library(vxu SHARED api/vxu.cpp)
 target_link_libraries(vxu openvx)
 
 if (OpenCL_FOUND)
+	add_definitions(-DCL_TARGET_OPENCL_VERSION=220)
```
kiritigowda, line 75: The definition assumes the version to be fixed, what happens when a version is defined or different? ( like in the windows flow. @paveltc can you run this PR on windows and Linux)

asalmanp: @kiritigowda It should still work because even without this change, OCL header is targeting the 220 version by default (i.e., the warning that shows up during the build), and I believe OpenVX apps have been working fine since this warning started to show up. Also, internally OpenVX checks the device version and, based on that targets specific features.

kiritigowda:
Using PR #413 for this PR

PR #406, opened 2020-12-10
Package update to support ROCm upgrade

pruthvistony:
- Package names are formed as per the conventions.
- ROCm version is appended to the version field when built by CI.
- make target 'package' tries building deb, rpm, zip and tz.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,11 @@
 # THE SOFTWARE.
 
 cmake_minimum_required(VERSION 3.0)
+
+find_program(MAKE_NSIS_EXE makensis)
+find_program(RPMBUILD_EXE rpmbuild)
+find_program(DPKG_EXE dpkg)
+
 project(MIVisionX)
 set(VERSION "1.9.4")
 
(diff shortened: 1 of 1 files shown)


kiritigowda (approved):
Approved. Pending @paveltc test results.

paveltc:
@kiritigowda The package installs after doing a 'sudo apt --fix-broken install' due to missing miopen-opencl.

pruthvistony:
> @kiritigowda The package installs after doing a 'sudo apt --fix-broken install' due to missing miopen-opencl.

Should the MIVisionX be allowed to install if miopen-opencl is NOT installed? If its so, need to remove the debian depends and rpm requires fields.

kiritigowda:
> > @kiritigowda The package installs after doing a 'sudo apt --fix-broken install' due to missing miopen-opencl.
> 
> Should the MIVisionX be allowed to install if miopen-opencl is NOT installed? If its so, need to remove the debian depends and rpm requires fields.

@pruthvistony Yes, MIVisionX has different components that can turn on or off depending on prereqs available on the system
With this approach, we can avoid having missing package errors.

Here is a table of MIVisionX prereqs and library- https://github.com/GPUOpen-ProfessionalCompute-Libraries/MIVisionX#docker-image

Let me know what you think, I will disable explicit package requirements in favor of implicit package dependencies.

kiritigowda:
Merging this PR to update the CMakeLists.

PR #407, opened 2020-12-11
Updated x265 build instruction to skip cmake GUI

rsudheerk001:
ccmake command from 'make-Makefiles.bash' opens a cmake GUI and halts the installation. 
replaced 'make-Makefiles.bash' with "cmake -G "Unix Makefiles" ../../source"


kiritigowda:
@rsudheerk001 can you sync to TOT MIVisionX to fix the conflicts.

PR #412, opened 2020-12-17
Setup - OpenCV with Display for RHEL

kiritigowda:
* Missing Libs install


kiritigowda:
Merging to get Dev-ops at speed

PR #438, opened 2021-02-01
fix for hang at the end for RALI app and training

rrawther:
Fixes the hang in image augmentation app. Also changed the python API to use original size as applicable.


kiritigowda (approved):

kiritigowda:
@paveltc LGTM, can you run the test?

paveltc:
@kiritigowda I'll run it and let you know.

paveltc:
@kiritigowda and @rrawther This fixes the hang I was seeing with the GPU unit tests.

PR #463, opened 2021-03-31
Dockers & CMake Updates

kiritigowda:
* Docker Files
  + CentOS 8 - Level 1 to 5
  + MIVisionX on CentOS
* RALI Deps Updates
* FFMPEG Find Fix
* CMake Find files updates


haampie:
Hi @kiritigowda, I think your FindOpenCL script is broken. It only finds OpenCL when CL_TARGET_OPENCL_VERSION is set, and that's ONLY set in the branch where you look for the rocm version of it.

For package managers it is an incorrect assumption to think everything is in /opt/rocm. As far as I can see, you can just as well remove the custom find script, and use cmake's script where you set `-DCMAKE_PREFIX_PATH=/opt/rocm` from the command line. 

Note that `CL_TARGET_OPENCL_VERSION` already defaults to 220 anyways in `cl_version.h`, and otherwise you can just enforce it in cmake...

kiritigowda:
@haampie thanks for taking the time to review this. The part of the code you are pointing was put in place to solve many of our user's requests. The problem being some installs had multiple versions of OpenCL installed and MIVisionX requires AMD's OpenCL extensions and we have to fall back on ROCm OpenCL. The ROCM_PATH is not hardcoded and can be set by the users to the point where it's installed.

Below is the issue pointed to `CL_TARGET_OPENCL_VERSION`
https://github.com/GPUOpen-ProfessionalCompute-Libraries/MIVisionX/issues/471#issue-856278211

If you still have questions or suggestions please contact me. Thanks!

haampie:
Hi @kiritigowda, there are multiple things incorrect with that script:

1. You are requiring `${ROCM_PATH}/opencl/lib/libOpenCL.so"` to exist, because if it does not, you never set CL_TARGET_OPENCL_VERSION, and you specify this as a requirement for your cmake script to find opencl.
  I don't have an `opencl` directory, so I cannot set `ROCM_PATH` in any sensible way. Please don't assume the user has an `opencl` directory in their install of rocm's opencl (e.g. when using spack install rocm-opencl or installing from the AUR).
2. You just assume the library is a shared lib, but the user may have a static version only. Please don't make assumptions about `.so` vs `.a`.

kiritigowda:
@haampie - the concerns of the install are clearly listed in the prerequisites section, these are questions or issues faced by people who do not read the instructions before installation. - Here is the prereqs section for [reference](https://github.com/GPUOpen-ProfessionalCompute-Libraries/MIVisionX#prerequisites)

We cover the majority of installations that follow the installation guide. For custom and installs that do not want to follow the prerequisites, the source code is available and the advanced users can make custom changes to match their needs.

We also welcome contributions to add functionality to allow these custom installations.  

@arjun-raj-kuppala as per this discussion you can add your requirements into the custom options to install @asalmanp can review and approve before accepting the changes.

PR #492, opened 2021-04-28
OpenVX HIP GPU backend - fix a bug in convolution kernels

asalmanp:


rrawther (approved):
please test before merging

PR #506, opened 2021-05-13
rocAL - Random BBox crop reader code clean up

fiona-gladwin:
Random BBox crop reader code clean up
Change USER_GIVEN_SIZE to MAX_SIZE for COCO partial decoding


rocAL/rocAL/source/randombboxcrop_reader.cpp
```diff
@@ -166,9 +149,9 @@ void RandomBBoxCropReader::read_all()
     bool invalid_bboxes;
     bool crop_success;
     BoundingBoxCord crop_box, jth_box;
+    uint32_t in_width, in_height;
```
rrawther, line 152: Please move these inside the while loop where it is used

kiritigowda (approved):

PR #524, opened 2021-06-09
SPACK Support - FindOpenCL - CL_TARGET_OPENCL_VERSION

kiritigowda:
* Allow required OpenCL version check without explicit path check


kiritigowda:
@rrawther & @asalmanp this PR is for SPACK fix. Without explicitly checking for ROCm OpenCL, we have to set CL version. Please review and provide any feedback.

asalmanp (approved):

PR #527, opened 2021-06-11
OpenVX HIP GPU backend - fix a bug for ColorKernels

asalmanp:
This PR fixes the color conversions for the following formats:
RGB->NV12, RGB->IYUV, UYVY->NV12, UYVY->IYUV, YUYV->NV12, YUYV->IYUV


rrawther (approved):
@kiritigowda : Looks like OpenVX conformance test is truncating the U and V values for 4:4:4 to 4:2:0 conversion, which caused us to fail the conformance since we were rounding it.
Can you check why this is done this way?

PR #534, opened 2021-06-18
H/canny gpu

hansely:


amd_openvx/openvx/api/vx_api.cpp
```diff
@@ -5688,7 +5688,7 @@ VX_API_ENTRY vx_threshold VX_API_CALL vxCreateThresholdForImage(vx_context conte
                     agoGenerateDataName(context, "thr", data->name);
                     agoAddData(&context->dataList, data);
 
-                    switch (output_format)
+                    switch (input_format)
```
asalmanp, line 5691: this change will cause new test failures for Threshold tests cases in CTS so this PR cannot be merged until the Threshold test cases are addressed.

asalmanp (changes requested):

PR #535, opened 2021-06-21
VxMapRemapPatch - Fix

kiritigowda:
Fix - Conditional jump on uninitialized value


kiritigowda:
@LakshmiKumar23 can you verify the fix for VxMapRemapPatch?

LakshmiKumar23:
> @LakshmiKumar23 can you verify the fix for VxMapRemapPatch?

@kiritigowda I still see the random fails. I ran it thrice, and it failed twice. The random fail is still seen

kiritigowda:
> > @LakshmiKumar23 can you verify the fix for VxMapRemapPatch?
> 
> @kiritigowda I still see the random fails. I ran it thrice, and it failed twice. The random fail is still seen

can you send me your run command?

kiritigowda:
@LakshmiKumar23 merging this to fix the Conditional jump on uninitialized value will look into remap patch again. Could not reproduce the failure after this fix.

PR #536, opened 2021-06-22
CTS Bug Fix for vxCanny & Support for int16 Threshold data type

hansely:
Check input image instead of output for the threshold data type. This fixes all the vxCanny CTS failures.
Added a support for int16 data type for threshold since it is now supported in OpenVX 1.3. This fixes all the vxThreshold CTS failures that were introduced by the above fix.


asalmanp (changes requested):
@hansely please also include the fix for Canny 7x7 for the HIP backend.

asalmanp (approved):

kiritigowda (approved):

PR #550, opened 2021-07-07
CI Updates - Tests

kiritigowda:
* Backend Build - HIP/OCL
* Tests - OCL + CPU
* Tests - HIP + CPU


kiritigowda:
@asalmanp this fix runs all HIP and OCL tests.

PR #564, opened 2021-07-13
rocal_pybind - fix for package link error

rrawther:
Fixes link issue where library is not found.

### DIFF
--- a/rocAL/rocAL_pybind/CMakeLists.txt
+++ b/rocAL/rocAL_pybind/CMakeLists.txt
@@ -58,7 +58,7 @@ endif()
 if(${BUILD_RALI_PYBIND})
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3")
 
-    link_directories(${ROCM_PATH}/rpp/lib)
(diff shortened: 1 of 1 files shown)


kiritigowda (approved):

rocAL/rocAL_pybind/CMakeLists.txt
```diff
@@ -58,7 +58,7 @@ endif()
 if(${BUILD_RALI_PYBIND})
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3")
 
-    link_directories(${ROCM_PATH}/rpp/lib)
+    link_directories(${ROCM_PATH}/rpp/lib ${ROCM_PATH}/mivisionx/lib)
```
asalmanp, line 61: @rrawther Which linking error does this PR try to fix, and why do we even need to link with mivisionx/lib folder? 

For a new build, if we don't install in the default path (i.e., /opt/rocm/mivisionx/), then this change is not working correctly, and it links with a previously installed mivisoinx wrongly (if there is an old mivisionx package installed).

rrawther:
@kiritigowda: Can you please merge these

PR #598, opened 2021-08-19
OpenVX GPU backend - fix a regression for HarrisCorner node introduced by PR#596

asalmanp:
PR#596 introduced a regression for HarrisCorner and it is causing a segfault for this node. PR#596 added an extra if check to set AGO_BUFFER_MERGE_FLAGS if the device_type is GPU but the issue is that the device_type is not initialized when we reach the ovxKernel_HarrisCorners function. 

This PR removes this extra if and it makes sure that the AGO_BUFFER_MERGE_FLAGS is unset if HarrisCorner is used in the graph.

### DIFF
--- a/amd_openvx/openvx/ago/ago_kernel_api.cpp
+++ b/amd_openvx/openvx/ago/ago_kernel_api.cpp
@@ -1868,9 +1868,7 @@ int ovxKernel_HarrisCorners(AgoNode * node, AgoKernelCommand cmd)
     //       use VX_KERNEL_AMD_HARRIS_SCORE_* kernels to compute Vc
     //       use VX_KERNEL_AMD_HARRIS_MERGE_SORT_AND_PICK_XY_HVC kernel for final step
     //       disable buffer merging for HarrisCorners
-    if(node->attr_affinity.device_type == AGO_KERNEL_FLAG_DEVICE_GPU){
-        agoSetEnvironmentVariable("AGO_BUFFER_MERGE_FLAGS", "1");
-    }
+    agoSetEnvironmentVariable("AGO_BUFFER_MERGE_FLAGS", "1");
     vx_status status = AGO_ERROR_KERNEL_NOT_IMPLEMENTED;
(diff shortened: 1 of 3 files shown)


kiritigowda (approved):

asalmanp:
started CI - http://math-ci.rocm.amd.com/blue/organizations/jenkins/compute-rocm-dkms-no-npi-hipclang%2FShort-GPUOpen%2FMIVisionX/detail/PR-598/2/pipeline

amd_openvx/openvx/ago/ago_kernel_api.cpp
```diff
@@ -1868,8 +1868,11 @@ int ovxKernel_HarrisCorners(AgoNode * node, AgoKernelCommand cmd)
     //       use VX_KERNEL_AMD_HARRIS_SCORE_* kernels to compute Vc
     //       use VX_KERNEL_AMD_HARRIS_MERGE_SORT_AND_PICK_XY_HVC kernel for final step
     //       disable buffer merging for HarrisCorners
-    if(node->attr_affinity.device_type == AGO_KERNEL_FLAG_DEVICE_GPU){
-        agoSetEnvironmentVariable("AGO_BUFFER_MERGE_FLAGS", "1");
+    char textBuffer[1024];
```
kiritigowda, line 1871: Can we enclose this workaround with `#if ENABLE_OPENCL` flag for both set and unset?

asalmanp: @kiritigowda added the guards for the set routine. The unset routine is inside the [agoGpuAllocBuffers](https://github.com/GPUOpen-ProfessionalCompute-Libraries/MIVisionX/blob/master/amd_openvx/openvx/ago/ago_drama_alloc.cpp#L63) function which already has the GPU guards.

kiritigowda (approved):

kiritigowda (approved):

PR #651, opened 2021-10-19
Neural Network Extension, HIP GPU backend - use object library for the HIP backend

AryanSalmanpour:


paveltc:
@asalmanp @rrawther @kiritigowda This PR passes all unit tests.

PR #708, opened 2021-12-09
update runVisionTests script to take image width and height as input arguments

asalmanp:


tests/vision_tests/runVisionTests.py
```diff
@@ -307,6 +317,14 @@ def strip_libtree_addresses(lib_tree):
     print("\nERROR: Vision Test Number of Frames not in range - [1 - 10000]\n")
     exit()
 
+if not 1 <= width <= 7680:
```
kiritigowda, line 320: Can we accommodate portrait mode -- w <= 4320 & h <= 7680 ?

kiritigowda (approved):

tests/vision_tests/runVisionTests.py
asalmanp: > Can we accommodate portrait mode -- w <= 4320 & h <= 7680 ?

yes, we can. I updated the PR to take images up to 7680x7680. So any image sizes within this range should be fine.

kiritigowda (approved):

PR #719, opened 2021-12-22
OpenVX - CMake Compiler Updates

kiritigowda:


kiritigowda:
@LakshmiKumar23 can you check this PR with cmake 3.22.X?

PR #742, opened 2022-01-26
Applications - Optical Flow for Object Tracking

kiritigowda:
* Track Objects
  + Live
  + Video


kiritigowda:
@LakshmiKumar23 @asalmanp - Test with [video](http://ewh.ieee.org/r6/scv/sps/openvx-material/PETS09-S1-L1-View001.avi)

kiritigowda:
Also use - https://github.com/GPUOpen-ProfessionalCompute-Libraries/MIVisionX/blob/master/data/videos/AMD_driving_virtual_20.mp4

PR #747, opened 2022-01-31
apps - update the dg_test and mivisionx_openvx_classifier apps to find the OpenCL package only if the OpenVX backend is OCL

asalmanp:


hansely:
Looks good on me

kiritigowda (approved):

PR #732, opened 2022-01-18
VX_NN - vxTensorCompare node implementation

hansely:
Implementation of vxTensorCompare node which is needed for mlperf models.
Supports the following comparison operation between two tensors: 
0 - Less than (<)
1 - Greater than (>)
2 - Less than or equal to (<=)
3 - Greater than or equal to (>=)
4 - Equal to (==)
5 - Not equal to (!=)

PR #762, opened 2022-02-07
VX_NN - bug fix

LakshmiKumar23:
Fixes a bug introduced by #732. Noticed by @paveltc and me while testing NN unit tests.
@paveltc Can you try building the C unit tests with this PR and let me know if the enum error goes away


paveltc:
@LakshmiKumar23 @kiritigowda This fixes the issue we were seeing.

PR #777, opened 2022-02-23
Tests - Video encoder test

paveltc:
Added amd_media_tests directory with encoder.gdf into MIVisionX/tests/


tests/amd_media_tests/README.md
```diff
@@ -0,0 +1,12 @@
+# AMD Media Tests
+
+Change the line in the encoder.gdf to contain full path:<br>
+data yuvimg  = image:1280,720,NV12:read,FULL-PATH/MIVisionX/tests/amd_media_tests/test.yuv
```
kiritigowda, line 4: can we use relative path like this - https://github.com/GPUOpen-ProfessionalCompute-Libraries/MIVisionX/blob/4faf7fb68688421a1d1f7a16d125cc8a1281f1cb/samples/gdf/canny.gdf#L6 ?

paveltc: @rrawther Rajy, is it possible to use relative paths here?

paveltc:
@kiritigowda @rrawther I got rid of the full path for the input as I found it is not needed with the encoder.

kiritigowda (approved):

PR #781, opened 2022-03-03
Develop - Merge into Master

kiritigowda:
* Readme Updates
* Tests Added
* rocAL Updates


paveltc:
@rrawther @kiritigowda I tried the decoder and encoder with this PR on a longer video clip and can get an mp4 output that plays in VLC media player.

PR #799, opened 2022-03-29
Docker - HIP Backend Fix

kiritigowda:
Docker - HIP Backend NN fix


kiritigowda:
CI does not test dockerfiles.

PR #812, opened 2022-04-12
OpenVX/rocAL - remove the call to hipDeviceReset

asalmanp:


LakshmiKumar23:
> @LakshmiKumar23 to verify it fixes the rocBLAS error in training

I do not see that rocBlas error either

PR #822, opened 2022-04-26
RunVX - suport for MIGraphX

LakshmiKumar23:
- Adding support for migraphx within runvx
- Adding a sample GDF to test


LakshmiKumar23:
Closing PR since migraphx node will be updated. Will create a new PR based on the new implementation

PR #831, opened 2022-05-04
rocAL - Fix for tar files present in a folder

swetha097:


rocAL/rocAL/source/label_reader_folders.cpp
```diff
@@ -142,14 +142,15 @@ void LabelReaderFolders::read_all(const std::string& _path)
         }
         else if(filesys::exists(pathObj) && filesys::is_directory(pathObj))
         {
+            dir_counter++;
             _folder_path = subfolder_path;
             _subfolder_file_names.clear();
             read_files(_folder_path);
             for(unsigned i = 0; i < _subfolder_file_names.size(); i++) {
-                add(_subfolder_file_names[i], dir_count);
+                add(_subfolder_file_names[i], dir_counter);
```
rrawther, line 150: I am not sure if this is the right fix for the issue. I think you want to increase the dir_count only for valid directories. The new code will put label =0 for all the subfolders which is not correct

rrawther (changes requested):
need to check the logic of this fix

rrawther (approved):

rrawther (commented):
@swetha097 : can you please remove all other changes in this PR except the one for the directory reading. Looks like it is not synced with tot

MIVisionX-setup.py
```diff
@@ -312,6 +312,12 @@
             os.system('sudo '+linuxFlag+' '+linuxSystemInstall +
                       ' '+linuxSystemInstall_check+' install -y miopen-hip migraphx')
 
+        #install MIGraphX
```
rrawther, line 315: @swetha097 can you please revert this change and sync with tot. This is not applicable to this PR

rrawther:
@kiritigowda: can we merge this PR

PR #840, opened 2022-05-13
 OpenVX migraphx node - modified the node to accept reversed dimensions for input/output tensors

asalmanp:


LakshmiKumar23:
@asalmanp Don't we have to change input dimensions order for resnet50 example as well?

LakshmiKumar23 (approved):

kiritigowda (approved):

Issue #837, opened 2022-05-11
Setup & Docker - Updates for NN Deps


kiritigowda:
`MIVisionX-Setup` & `Docker files` need updates to install NN deps with ROCm `5.1.1+`

kiritigowda:
PR #841 solves this issue

PR #841, opened 2022-05-13
NN Deps - Docker & Setup Updates

kiritigowda:
* Docker - Use released packages for miopen 
* Setup - Build from source removed for MIOpen & deps


kiritigowda:
Fixes issue #837

PR #853, opened 2022-06-06
[CMAKE] GNUInstall Dir enhancements

arvindcheru:
Summary of proposed changes:

- Updated GNU install dir enhancements


Mystro256:
Looks good

asalmanp (approved):

CMakeLists.txt
```diff
@@ -123,7 +126,7 @@ else()
 endif()
 
 # install MIVisionX docs -- {ROCM_PATH}/share/doc/mivisionx/
-install(DIRECTORY docs DESTINATION share/doc/mivisionx)
+install(DIRECTORY docs DESTINATION ${CMAKE_INSTALL_DATADIR}/doc/mivisionx)
```
nunnikri, line 129: We need to replace this with DOCDIR instead of DATADIR

CMakeLists.txt
```diff
@@ -148,7 +151,7 @@ set(CPACK_RPM_PACKAGE_SUMMARY      "AMD MIVisionX is a comprehensive Computer Vi
 # set license information
 set(CPACK_RPM_PACKAGE_LICENSE    "MIT")
 set(CPACK_RESOURCE_FILE_LICENSE  "${CMAKE_CURRENT_SOURCE_DIR}/LICENSE.txt")
-install(FILES ${CPACK_RESOURCE_FILE_LICENSE} DESTINATION share/doc/mivisionx)
+install(FILES ${CPACK_RESOURCE_FILE_LICENSE} DESTINATION ${CMAKE_INSTALL_DATADIR}/doc/mivisionx)
```
nunnikri, line 154: We need to replace this with DOCDIR instead of DATADIR

nunnikri (changes requested):

kiritigowda:
@arvindcheru this pr has merge conflicts.

amd_openvx/openvx/CMakeLists.txt
```diff
@@ -21,6 +21,11 @@
 cmake_minimum_required(VERSION 3.0)
 project(openvx)
 
+# Default libdir to "lib", this skips GNUInstallDirs from trying to take a guess if it's unset:
+set(CMAKE_INSTALL_LIBDIR "lib" CACHE STRING "Library install directory")
+#including GNUInstall Dir definitions.
+include( GNUInstallDirs )
```
kiritigowda, line 27: do we need this - `include( GNUInstallDirs )` in every file or can we just cache it in the main file and let it stream down to all?

CMakeLists.txt
```diff
@@ -22,6 +22,9 @@ cmake_minimum_required(VERSION 3.0)
 project(MIVisionX)
 set(VERSION "2.2.0")
 
+#including GNUInstall Dir definitions.
+include( GNUInstallDirs )
+
 set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
```
kiritigowda, line 28: Needs update -- set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR})

CMakeLists.txt
```diff
@@ -22,6 +22,9 @@ cmake_minimum_required(VERSION 3.0)
...
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
```
kiritigowda, line 30: CMAKE_INSTALL_BINDIR

kiritigowda:
@arvindcheru requested a few changes. Please let me know what you think.

CMakeLists.txt
arvindcheru: as discussed since the project name is difference from the folder name, not added this right now.

CMakeLists.txt
```diff
@@ -22,12 +22,18 @@ cmake_minimum_required(VERSION 3.0)
...
+  set(${CMAKE_INSTALL_LIBDIR} "lib")
+  set(${CMAKE_INSTALL_DATADIR} "share")
+  set(${CMAKE_INSTALL_INCLUDEDIR} "include")
+  set(${CMAKE_INSTALL_BINDIR} "bin")
+endif()
 
-set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR})
```
arvindcheru, line 34: install path destination we were only updating.

CMakeLists.txt
```diff
@@ -40,7 +46,14 @@ if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
 endif(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 
-option(ENHANCED_MESSAGE "MIVisionX Enhanced Message Option" ON)
+# MIVisionX Default Options
```
arvindcheru, line 49: Do we require this change to be added along with?

CMakeLists.txt
```diff
@@ -22,12 +22,18 @@ cmake_minimum_required(VERSION 3.0)
 project(MIVisionX)
 set(VERSION "2.2.0")
 
-#including GNUInstall Dir definitions.
-include( GNUInstallDirs )
+if(UNIX)
```
arvindcheru, line 25: Thanks for the changes, will check & verify build removing other includes.

CMakeLists.txt
```diff
@@ -22,12 +22,18 @@ cmake_minimum_required(VERSION 3.0)
...
+  include(GNUInstallDirs CACHE)
+else()
+  set(${CMAKE_INSTALL_LIBDIR} "lib")
```
lawruble13, line 28: I'm not certain about this at a high level, and at a technical level it's currently incorrect. This will dereference the variable `CMAKE_INSTALL_LIBDIR` and use it's value as a variable name to set. This should be `set(CMAKE_INSTALL_LIBDIR "lib")`, but I'm not even sure this is something we should be doing.

amd_openvx/openvx/CMakeLists.txt
arvindcheru: Removed all subdir level inclusions in latest version

CMakeLists.txt
kiritigowda: @lawruble13 the Unix will include GNU Dirs in  Linux & macOS. What do we do in WIN32?

CMakeLists.txt
```diff
@@ -22,9 +22,18 @@ cmake_minimum_required(VERSION 3.0)
...
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
+if(UNIX)
+  include(GNUInstallDirs CACHE)
+else()
+  set(CMAKE_INSTALL_LIBDIR "lib")
+  set(CMAKE_INSTALL_DATADIR "share")
+  set(CMAKE_INSTALL_INCLUDEDIR "include")
+  set(CMAKE_INSTALL_BINDIR "bin")
+endif()
```
cgmb, line 32: I'm not sure I follow why there's this if/else. Those are the directory names that GNUInstallDirs would be using on Windows anyway. On Linux, without setting a default for the `CMAKE_INSTALL_LIBDIR`, you're going to end up with your libraries in `/opt/rocm/lib64` on RHEL and SLES, which I don't think is the intent.

I think @arvindcheru had it right. This whole if/else should just be:
```cmake
set(CMAKE_INSTALL_LIBDIR "lib" CACHE STRING "Library install directory")
include(GNUInstallDirs)
```

cgmb (approved):
This looks reasonable to me.

kiritigowda (approved):

PR #854, opened 2022-06-06
rocAL Unit Test - Fix & Updates

kiritigowda:
* OpenCV 3/4 Support
* Cleanup


kiritigowda:
@paveltc -- this should fix OpenCV `3`/`4 ` support and API change fix. Can you verify?

PR #932, opened 2022-08-30
define vx_float16 data type

rrawther:
Defining vx_float16 type for tensor data type

### DIFF
--- a/amd_openvx/openvx/include/vx_ext_amd.h
+++ b/amd_openvx/openvx/include/vx_ext_amd.h
@@ -30,6 +30,13 @@ THE SOFTWARE.
 #include <string>
 #endif
 
+#ifndef EXPERIMENTAL_PLATFORM_SUPPORTS_16_FLOAT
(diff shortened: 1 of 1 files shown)


kiritigowda (commented):
@rrawther This is defined in vx types -- https://github.com/KhronosGroup/OpenVX-api-docs/blob/b6de9eda2c2b1590a3762ce4596e7aa8f092bb83/include/VX/vx_types.h#L119

Did you want to the headers to use half ?

rrawther:
@kiritigowda : As I mentioned the type in vx_types.h is for supporting some special hardware architectures. For x86 CPUs and GPUs, we need to use half library which defines the IEEE half data type which can be interpreted as vx_float16 data type. 
Can you please look into the compile error which says half.h is not found. Not sure why as it is building in my system

rrawther:
@kiritigowda confirmed that it will get fixed in OpenVX

PR #948, opened 2022-09-14
SPACK updates - Turbo JPEG 

kiritigowda:
@srekolam these updates add an ENV Variable - `TURBO_JPEG_PATH` that could be set or pass `TURBO_JPEG_PATH` as a -D cmake build option. Let me know if this would fix SPACK build.

### DIFF
--- a/amd_openvx/cmake/FindTurboJpeg.cmake
+++ b/amd_openvx/cmake/FindTurboJpeg.cmake
@@ -25,21 +25,30 @@
 ################################################################################
 find_path(TurboJpeg_INCLUDE_DIRS
     NAMES turbojpeg.h
+    HINTS
+    $ENV{TURBO_JPEG_PATH}/include
     PATHS
+    ${TURBO_JPEG_PATH}/include
     /usr/include/
 )
 mark_as_advanced( TurboJpeg_INCLUDE_DIRS )
 
 find_library( TurboJpeg_LIBRARIES
     NAMES libturbojpeg.so
+    HINTS
+    $ENV{TURBO_JPEG_PATH}/lib
     PATHS
+    ${TURBO_JPEG_PATH}/include
     /usr/lib
 )
 mark_as_advanced( TurboJpeg_LIBRARIES_DIR )
 
 find_path(TurboJpeg_LIBRARIES_DIR
     NAMES libturbojpeg.so
+    HINTS
+    $ENV{TURBO_JPEG_PATH}/lib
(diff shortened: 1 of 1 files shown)


amd_openvx/cmake/FindTurboJpeg.cmake
```diff
@@ -25,21 +25,30 @@
...
     /usr/lib
 )
 mark_as_advanced( TurboJpeg_LIBRARIES_DIR )
 
 find_path(TurboJpeg_LIBRARIES_DIR
     NAMES libturbojpeg.so
+    HINTS
+    $ENV{TURBO_JPEG_PATH}/lib
     PATHS
+    ${TURBO_JPEG_PATH}/include
```
srekolam, line 51: ${TURBO_JPEG_PATH}/lib

amd_openvx/cmake/FindTurboJpeg.cmake
```diff
@@ -25,21 +25,30 @@
...
     /usr/include/
 )
 mark_as_advanced( TurboJpeg_INCLUDE_DIRS )
 
 find_library( TurboJpeg_LIBRARIES
     NAMES libturbojpeg.so
+    HINTS
+    $ENV{TURBO_JPEG_PATH}/lib
     PATHS
+    ${TURBO_JPEG_PATH}/include
```
srekolam, line 41: ${TURBO_JPEG_PATH}/lib

srekolam (changes requested):
i am assuming there is typo with respect to find_library() you have included ${TURBO_JPEG_PATH}/include instead of ${TURBO_JPEG_PATH}/lib

amd_openvx/cmake/FindTurboJpeg.cmake
srekolam: also, it would be good to add lib64 too...as when i tried with this change , i found the libjpegturbo.so was installed in /lib64  as in the case below 
./opt/spack/linux-centos8-skylake_avx512/gcc-8.4.1/libjpeg-turbo-2.1.3-y5dxc36owqbi5kxgk6jih7543nbmvbkw/lib64/libturbojpeg.so

kiritigowda: @srekolam added support for lib64

srekolam (commented):
LGTM!! Thanks

asalmanp (approved):

PR #950, opened 2022-09-16
rocAL - Resize scaling modes support

fiona-gladwin:
- Add interpolation type, scaling mode, resize shorter, resize longer and max size parameters to resize C++ and python API
- Add support for following scaling modes:
  1. Default
  2. Stretch
  3. Not smaller
  4. Not larger


rocAL/rocAL/include/api/rocal_api_augmentation.h
```diff
@@ -45,10 +45,20 @@ extern "C"  RocalImage  ROCAL_API_CALL rocalSequenceRearrange(RocalContext conte
...
+/// \param interpolation_type The type of interpolation to be used for resize.
 /// \return
 extern "C"  RocalImage  ROCAL_API_CALL rocalResize(RocalContext context, RocalImage input,
-                                                unsigned dest_width, unsigned dest_height,
-                                                bool is_output );
+                                                   unsigned dest_width, unsigned dest_height,
+                                                   bool is_output,
+                                                   RocalResizeScalingMode scaling_mode = ROCAL_SCALING_MODE_STRETCH,
+                                                   std::vector<unsigned> max_size = {},
+                                                   unsigned resize_shorter = 0,
```
rrawther, line 59: instead of passing 2 separate parameters for resize_shorted and resize_longer, we can use a bitmask where each bit can represent the flag. That way it is easy to extend for future modes and reduces an argument

rocAL/rocAL/include/augmentations/geometry_augmentations/node_resize.h
```diff
@@ -22,19 +22,27 @@ THE SOFTWARE.
...
     vx_array get_src_height() { return _src_roi_height; }
+    void init(unsigned dest_width, unsigned dest_height, RocalResizeScalingMode scaling_mode,
+              std::vector<unsigned> max_size, RocalResizeInterpolationType interpolation_type);
+    void adjust_out_roi_size();
 protected:
     void create_node() override;
     void update_node() override;
 private:
-    vx_array  _dst_roi_width , _dst_roi_height ;
+    vx_array  _dst_roi_width , _dst_roi_height;
```
rrawther, line 42: we need to get rid of roi arrays and use single tensor for ROIs when we have tensor implementation

rocAL/rocAL/source/api/rocal_api_augmentation.cpp
```diff
@@ -66,6 +66,24 @@ THE SOFTWARE.
 #include "commons.h"
 #include "context.h"
 #include "rocal_api.h"
+#include "image_source_evaluator.h"
+
+void
+get_max_resize_width_and_height(std::vector<unsigned> &src_size, std::vector<unsigned> &dst_size, RocalResizeScalingMode mode,
+                                std::vector<unsigned> &out_size) {
+    float max_aspect_ratio = 3.0f;  // assuming maximum aspect ratio for an image as 3.0
```
rrawther, line 74: please use a #define for MAX_ASPECT_RATIO, so it is not hardcoded within code

rocAL/rocAL/include/augmentations/geometry_augmentations/node_resize.h
```diff
@@ -22,19 +22,27 @@ THE SOFTWARE.
...
+    unsigned _dest_width, _dest_height;
+    int _interpolation_type;
+    RocalResizeScalingMode _scaling_mode;
+    std::vector<float> _src_roi_size;
```
rrawther, line 46: Is there any reason to keep the src_roi in flat and dst_roi in uint?

rocAL/rocAL/source/api/rocal_api_augmentation.cpp
```diff
@@ -517,19 +541,69 @@ rocalResize(
     auto input = static_cast<Image*>(p_input);
     try
     {
-        // For the resize node, user can create an image with a different width and height
+        if((dest_width || dest_height || resize_longer || resize_shorter) == 0)
```
rrawther, line 544: please use | instead of || so compiler will use logical or

rocAL/rocAL/source/api/rocal_api_augmentation.cpp
```diff
@@ -517,19 +541,69 @@ rocalResize(
...
+            THROW("Atleast one size 'dest_width' or 'dest_height' or 'resize_shorter' or 'resize_longer' must be specified")
+        if((dest_width || dest_height) && (resize_longer || resize_shorter))
+            THROW("Only one method of specifying size can be used \ndest_width and/or dest_height\nresize_shorter\nresize_longer")
+        if(resize_longer && resize_shorter)
+            THROW("'resize_longer' and 'resize_shorter' cannot be passed together. They are mutually exclusive.")
+
         ImageInfo output_info = input->info();
-        if (dest_width == 0) dest_width = input->info().width();
-        if (dest_height == 0) dest_height = input->info().height_single();
+        unsigned dst_width, dst_height;
```
rrawther, line 552: can you rename it as out_width and out_height. Confusing with dest_width

rocAL/rocAL/source/api/rocal_api_augmentation.cpp
```diff
@@ -505,7 +523,13 @@ rocalResize(
...
         unsigned dest_height,
-        bool is_output)
+        bool is_output,
+        RocalResizeScalingMode scaling_mode,
+        std::vector<unsigned> max_size,
+        unsigned resize_shorter,
+        unsigned resize_longer,
+        RocalResizeInterpolationType interpolation_type
+)
 {
```
rrawther, line 533: move to previous line to follow google coding style

rocAL/rocAL/source/api/rocal_api_augmentation.cpp
```diff
@@ -517,19 +541,69 @@ rocalResize(
...
+            resize_scaling_mode = scaling_mode;
+            dst_width = dest_width;
+            dst_height = dest_height;
+        }
 
-        output_info.width(dest_width);
-        output_info.height(dest_height);
+        std::vector<unsigned> maximum_size;
+        if (max_size.size()) {
+            if(max_size.size() == 1) {
```
rrawther, line 570: can't you specify height only for max_size with one dimension?

rocAL/rocAL/source/api/rocal_api_augmentation.cpp
```diff
@@ -517,19 +541,69 @@ rocalResize(
...
+        std::vector<unsigned> output_info_size(2, 0);
+        if (maximum_size.size() && maximum_size[0] != 0 && maximum_size[1] != 0) {
+            // If max_size is passed by the user, the resized images cannot exceed the max size,
+            output_info_size = maximum_size;
+        } else {
+            std::vector<unsigned> src_size = {input->info().width(), input->info().height_single()};
+            std::vector<unsigned> dst_size = {dst_width, dst_height};
+
+            // compute the output info width and height wrt the scaling modes and roi passed
+            get_max_resize_width_and_height(src_size, dst_size, resize_scaling_mode, output_info_size);
```
rrawther, line 589: There is only few lines of code inside the fun get_max_resize_width_and_height. It is better to replace that code here and get rid of the function

rocAL/rocAL/source/augmentations/geometry_augmentations/node_resize.cpp
```diff
@@ -54,10 +54,114 @@ void ResizeNode::create_node()
...
+    _dest_height = dest_height;
+    _interpolation_type = (int)interpolation_type;
+    _src_roi_size.resize(2);
+    _dst_roi_size.resize(2);
+    _max_roi_size = max_size;
+}
+
+void ResizeNode::adjust_out_roi_size() {
+    const unsigned dim = 2; // Currently supports only 2D images
+    std::vector<double> scale(dim, 1);
```
rrawther, line 97: do we need to use double for scale instead of float?

rocAL/rocAL/source/augmentations/geometry_augmentations/node_resize.cpp
```diff
@@ -54,10 +54,114 @@ void ResizeNode::create_node()
...
+    std::vector<bool> has_size(dim, false);
+    unsigned sizes_provided = 0;
+    bool has_max_size = _max_roi_size.size() > 0;
+    for (unsigned i = 0; i < dim; i++) {
+        has_size[i] = (_src_roi_size[i] != 0) && (_dst_roi_size[i] != 0);
+        sizes_provided += has_size[i];
+        scale[i] = _src_roi_size[i] ? (_dst_roi_size[i] / static_cast<double>(_src_roi_size[i])) : 1;
```
rrawther, line 104: please cast the numerator instead

rocAL/rocAL/source/augmentations/geometry_augmentations/node_resize.cpp
```diff
@@ -54,10 +54,114 @@ void ResizeNode::create_node()
...
+            }
+        }
+    } else if (_scaling_mode == RocalResizeScalingMode::ROCAL_SCALING_MODE_DEFAULT) {
+        if (sizes_provided < dim) {
+            double average_scale = 1;
+            for (unsigned i = 0; i < dim; i++) {
+                if (has_size[i])
+                    average_scale *= scale[i];
+            }
+            if (sizes_provided > 1)
```
rrawther, line 126: There is something wrong about this logic where average_scale is computed twice

rocAL/rocAL/source/augmentations/geometry_augmentations/node_resize.cpp
rrawther, line 97: please use an array of 2 instead of vector here

rocAL/rocAL/source/augmentations/geometry_augmentations/node_resize.cpp
```diff
@@ -54,10 +54,114 @@ void ResizeNode::create_node()
...
+                average_scale = std::pow(average_scale, 1.0 / sizes_provided);
```
rrawther, line 127: what is the purpose of this line?

rocAL/rocAL/source/augmentations/geometry_augmentations/node_resize.cpp
```diff
@@ -54,10 +54,114 @@ void ResizeNode::create_node()
...
+            for(unsigned i = 0; i < dim; i++) {
+                if(!has_size[i])
+                    _dst_roi_size[i] = std::round(_src_roi_size[i] * average_scale);
```
rrawther, line 130: this is true dst_roi_size is not specified

rocAL/rocAL/source/augmentations/geometry_augmentations/node_resize.cpp
```diff
@@ -54,10 +54,114 @@ void ResizeNode::create_node()
...
+        if (has_max_size) {
+            for (unsigned i = 0; i < dim; i++) {
+                if ((_max_roi_size[i] > 0) && (_dst_roi_size[i] > _max_roi_size[i]))
+                    _dst_roi_size[i] = _max_roi_size[i];
+            }
+        }
+    } else {
+        double final_scale = 0;
+        bool first = true;
+        for (unsigned i = 0; i < dim; i++) {
```
rrawther, line 142: this logic is very ambiguous. Please use std::min and std::max of 2 scale values instead

rocAL/rocAL/source/augmentations/geometry_augmentations/node_resize.cpp
```diff
@@ -54,10 +54,114 @@ void ResizeNode::create_node()
...
+    _dest_width = dest_width;
+    _dest_height = dest_height;
+    _interpolation_type = (int)interpolation_type;
+    _src_roi_size.resize(2);
+    _dst_roi_size.resize(2);
+    _max_roi_size = max_size;
+}
+
+void ResizeNode::adjust_out_roi_size() {
+    const unsigned dim = 2; // Currently supports only 2D images
```
rrawther, line 96: the entire logic can be much simplified if we just get rid of dim loops assume const dimension of 2.

rrawther (changes requested):
@fiona-gladwin : please address review  comments

rocAL/rocAL/include/api/rocal_api_augmentation.h
fiona-gladwin: RPP accepts resize width and height as unsigned integers. If we change the type of dst_width(resize_x), dest_height(resize_y), resize_shorter and resize_longer to float in order to match with DALI resize API we may need to change the way RPP accepts the inputs too

rocAL/rocAL/source/augmentations/geometry_augmentations/node_resize.cpp
fiona-gladwin: The average scale computation comes into picture when we have more than 2 dimensions, Since we are going to restrict our code to only width and height (i.e 2-dimension) we can eliminate the average scale computation

rocAL/rocAL/include/augmentations/geometry_augmentations/node_resize.h
fiona-gladwin: In future if we fuse resize with ROI similar to DALI API we may get the src width and height values in float when relative roi calculations comes into picture, hence the src_roi was kept as float whereas RPP accepts the resize width and height values as uint

rocAL/rocAL/source/augmentations/geometry_augmentations/node_resize.cpp
fiona-gladwin: In future if we fuse resize with ROI similar to DALI API we will have src width and height values (in float), using double we were able to resolve the rounding off errors

rocAL/rocAL/source/augmentations/geometry_augmentations/node_resize.cpp
rrawther: I meant to remove the dim loop for all the code. But I still see lots of places it is used.

fiona-gladwin:
@rrawther We have addressed all the review comments and made the necessary changes, Please check

rocAL/rocAL/source/augmentations/geometry_augmentations/node_resize.cpp
```diff
@@ -54,10 +53,82 @@ void ResizeNode::create_node()
...
-{
-  
-}
\ No newline at end of file
+void ResizeNode::init(unsigned dest_width, unsigned dest_height, RocalResizeScalingMode scaling_mode,
+                      std::vector<unsigned> max_size, RocalResizeInterpolationType interpolation_type) {
+    _scaling_mode = scaling_mode;
+    _out_width = dest_width;
+    _out_height = dest_height;
+    if(max_size.size() > 0) {
```
rrawther, line 88: what is the value of max_width and max_height for else clause?

rocAL/rocAL/source/augmentations/geometry_augmentations/node_resize.cpp
```diff
@@ -54,10 +53,82 @@ void ResizeNode::create_node()
...
+        if (_dst_width == 0) _dst_width = _src_width;
+        if (_dst_height == 0) _dst_height = _src_height;
+
+        if (has_max_size) {
+            if (_max_width != 0) _dst_width = std::min(_dst_width, _max_width);
+            if (_max_height != 0) _dst_height = std::min(_dst_height, _max_height);
+        }
+    } else if (_scaling_mode == RocalResizeScalingMode::ROCAL_SCALING_MODE_DEFAULT) {
+        if (_dst_width == 0 && _dst_height != 0) {  // Only height is passed
+            _dst_width = static_cast<uint>(std::round(_src_width * (static_cast<float>(_dst_height) / _src_height)));
```
rrawther, line 107: you can use std::lround instead of std::round plus cast

rrawther (changes requested):
@fiona-gladwin: please address review comments

rocAL/rocAL/source/augmentations/geometry_augmentations/node_resize.cpp
fiona-gladwin: Only if user specifies the value the max_width and max_height it will be set, Otherwise it would be zero and not taken into consideration

fiona-gladwin:
> @fiona-gladwin: please address review comments

@rrawther I have addressed the review comments please check

rrawther (approved):
@LakshmiKumar23 to verify it is working

rocAL/rocAL/source/augmentations/geometry_augmentations/node_resize.cpp
rrawther: If max_size is a read_only parameter, please pass by reference with const type

rocAL/rocAL/source/augmentations/geometry_augmentations/node_resize.cpp
```diff
@@ -54,10 +53,83 @@ void ResizeNode::create_node()
...
+
+void ResizeNode::adjust_out_roi_size() {
+    bool has_max_size = (_max_width | _max_height) > 0;
+
+    if (_scaling_mode == RocalResizeScalingMode::ROCAL_SCALING_MODE_STRETCH) {
+        if (_dst_width == 0) _dst_width = _src_width;
+        if (_dst_height == 0) _dst_height = _src_height;
+
+        if (has_max_size) {
+            if (_max_width != 0) _dst_width = std::min(_dst_width, _max_width);
```
rrawther, line 103: if (_max_width) is better

rocAL/rocAL/source/augmentations/geometry_augmentations/node_resize.cpp
```diff
@@ -54,10 +53,83 @@ void ResizeNode::create_node()
...
+            if (_max_height != 0) _dst_height = std::min(_dst_height, _max_height);
+        }
+    } else if (_scaling_mode == RocalResizeScalingMode::ROCAL_SCALING_MODE_DEFAULT) {
+        if (_dst_width == 0 && _dst_height != 0) {  // Only height is passed
```
rrawther, line 107: if you use logical statements instead, compiler generates better code than conditional statements
for e.g., if (!_dst_width & _dst_height)

rrawther (commented):
@LakshmiKumar23  to verify

rocAL/rocAL/source/augmentations/geometry_augmentations/node_resize.cpp
fiona-gladwin: Done

rocAL/rocAL/source/augmentations/geometry_augmentations/node_resize.cpp
fiona-gladwin: Done

kiritigowda:
@LakshmiKumar23 is this ready to be merged?

LakshmiKumar23:
> @LakshmiKumar23 is this ready to be merged?

@kiritigowda Yes this can be merged

PR #956, opened 2022-09-22
AMD - Float16 Support

kiritigowda:
OpenVX defines float16 with Experimental flag using hfloat - https://github.com/GPUOpen-ProfessionalCompute-Libraries/MIVisionX/blob/9b1c5c487722c8f09ab901727def22e78549e0b6/amd_openvx/openvx/include/VX/vx_types.h#L114

To use C++ half.hpp - turn on `AMD_FP16_SUPPORT`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,7 @@ option(ROCAL            "Build MIVisionX with ROCAL Support"       ON)
 option(LOOM             "Build MIVisionX with LOOM Support"        ON)
 option(GPU_SUPPORT      "Build MIVisionX with GPU Support"         ON)
 option(MIGRAPHX         "Build MIVisionX with MIGraphX Support"    ON)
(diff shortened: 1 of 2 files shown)


LakshmiKumar23:
@kiritigowda we might have to give this as an option to set in the cmake so that user can set it.
Shows this error otherwise:
CMake Warning:
  Manually-specified variables were not used by the project:
  AMD_FLOAT_16_SUPPORT

kiritigowda:
> @kiritigowda we might have to give this as an option to set in the cmake so that user can set it.
> Shows this error otherwise:
> CMake Warning:
> Manually-specified variables were not used by the project:
> AMD_FLOAT_16_SUPPORT

Change added

LakshmiKumar23 (approved):
@kiritigowda Builds successfully. @fiona-gladwin also confirmed with tensor PR. Can be merged

PR #962, opened 2022-09-28
AMD Media Decoder - measure decode time

paveltc:
Decode time can now be printed out for the AMD Media Decoder.
Need to set ENABLE_PERF_MEASURE to 1 in decoder.cpp in order to turn on the timing and printing.

### DIFF
--- a/amd_openvx_extensions/amd_media/decoder.cpp
+++ b/amd_openvx_extensions/amd_media/decoder.cpp
@@ -34,6 +34,9 @@ THE SOFTWARE.
 #include <sstream>
 #include <stdlib.h>
 
+// Performance measurement option
+#define ENABLE_PERF_MEASURE        0
+
 // OpenCL configuration
 #define DUMP_DECODED_FRAME         0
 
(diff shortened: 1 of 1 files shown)


amd_openvx_extensions/amd_media/decoder.cpp
```diff
@@ -741,6 +754,11 @@ void CLoomIoMediaDecoder::DecodeLoop(int mediaIndex)
                 return;
             }
             int status = avcodec_receive_frame(videoCodecContext[mediaIndex], frame);
+#if ENABLE_PERF_MEASURE
```
asalmanp, line 757: remove this code (line 757-761) and put it after this [line](https://github.com/GPUOpen-ProfessionalCompute-Libraries/MIVisionX/blob/master/amd_openvx_extensions/amd_media/decoder.cpp#L757)

amd_openvx_extensions/amd_media/decoder.cpp
```diff
@@ -873,6 +891,9 @@ void CLoomIoMediaDecoder::DecodeLoop(int mediaIndex)
             decodeFrameCount[mediaIndex]++;
             PushAck(mediaIndex, 0);
         }
+#if ENABLE_PERF_MEASURE
```
asalmanp, line 894: remove this code line (894-896) and put after this [line](https://github.com/GPUOpen-ProfessionalCompute-Libraries/MIVisionX/blob/master/amd_openvx_extensions/amd_media/decoder.cpp#L725) to print the final average time

asalmanp (changes requested):

PR #975, opened 2022-10-13
rocAL - python unit test update

LakshmiKumar23:
Updates the way python version is found to be compatible with python3.10
Old version accepts only single digits ```pythonX.Y``` for instance not ```pythonX.YW```


rrawther (approved):

rrawther:
@LakshmiKumar23: The same change needs to be done for READERS_TEST_FILE.sh as well

PR #976, opened 2022-10-14
[Vision tests] - add a new gdf test for scaling an NV12 image and converting it to an RGB image

asalmanp:


paveltc (approved):
I have run this GDF and it runs fine. Ready to merge.

PR #994, opened 2022-11-10
OS Support - Updates

kiritigowda:
* Docker Updates
* Setup Updates
* OpenCV Upgrade - 4.6.0


MIVisionX-setup.py
```diff
@@ -362,7 +364,7 @@
                   ' install -t pattern devel_basis')
         os.system('sudo '+linuxFlag+' '+linuxSystemInstall+' '+linuxSystemInstall_check +
                   ' install gtk2-devel libjpeg-devel libpng-devel libtiff-devel libavc1394 wget unzip')
-    # OpenCV 4.5.5
+    # OpenCV 4.6.0
```
LakshmiKumar23, line 367: @kiritigowda  Seems like you need this flag "-D WITH_GTK=ON" to build OpenCV

LakshmiKumar23: @kiritigowda Please add these two flags too for jpeg opencv 
"-D WITH_JPEG=ON -D BUILD_JPEG=ON"

LakshmiKumar23 (changes requested):

MIVisionX-setup.py
LakshmiKumar23: @kiritigowda Please add these two flags to the dockerfiles also. I tested with ub20 & ub22 level 3 and it works with jpg images and opencv4.6

PR #997, opened 2022-11-22
AMD OpenVX - HIP cleanup

rrawther:
removed hipStream creation from OpenVX context since it is not used


amd_openvx/openvx/ago/ago_util_hip.cpp
```diff
@@ -98,14 +84,13 @@ int agoGpuHipCreateContext(AgoContext *context, int deviceID) {
...
 int agoGpuHipReleaseContext(AgoContext * context) {
-    if (context->hip_stream) {
-        hipError_t status = hipStreamDestroy(context->hip_stream);
-        if (status != hipSuccess) {
-            agoAddLogEntry(NULL, VX_FAILURE, "ERROR: agoGpuHipReleaseContext: hipStreamDestroy(%p) failed (%d)\n", context->hip_stream, status);
-            return -1;
-        }
-        context->hip_stream = NULL;
+
+    hipError_t status = hipDeviceReset();
```
asalmanp, line 88: @rrawther we shouldn't call the hipDeviceReset here. It's not safe to call here as it deletes all the resources associated with the current device.

You would need to call the hipStreamDestroy for each stream created for each Graph to delete them.

asalmanp (changes requested):

asalmanp (approved):

PR #1005, opened 2022-12-02
rocAL : Fix build errors wrt Boost version:1.80 and build warnings in Ubuntu version:22

fiona-gladwin:


rocAL/rocAL/source/meta_data/caffe2_meta_data_reader.cpp
```diff
@@ -113,20 +113,20 @@ void Caffe2MetaDataReader::read_lmdb_record(std::string file_name, uint file_byt
     string str_key;
 
     // Creating an LMDB environment handle
-	E(mdb_env_create(&env));
+	CHECK_LMDB(mdb_env_create(&env));
```
LakshmiKumar23, line 116: @fiona-gladwin Can you check formatting on this file?

LakshmiKumar23:
@kiritigowda can be merged once Fiona fixes the formatting

rocAL/rocAL/include/readers/image/reader.h
```diff
@@ -27,10 +27,12 @@ THE SOFTWARE.
 #include <tuple>
 #include "meta_data_reader.h"
 
-#define E(expr) CHECK_CAFFE((rc = (expr)) == MDB_SUCCESS, #expr)
-#define CHECK_CAFFE(test, msg) \
-    ;                          \
-    ((test) ? (void)0 : ((void)fprintf(stderr, "%s:%d: %s: %s\n", __FILE__, __LINE__, msg, mdb_strerror(rc)), abort()))
+#define CHECK_LMDB(status)          \
+    do {                            \
```
rrawther, line 31: Why do we need a do{ } while loop here? Isn't empty braces enough?

rocAL/rocAL/source/meta_data/caffe2_meta_data_reader.cpp
rrawther, line 116: Please use more meaningful name like ERROR_LMDB_RETURN_STATUS()

rrawther (changes requested):
please address review comments

rocAL/rocAL/include/readers/image/reader.h
fiona-gladwin: This macro accepts the value returned by the [LMDB API](http://www.lmdb.tech/doc/group__mdb.html#gaad6be3d8dcd4ea01f8df436f41d158d4), which is an int, checks if it is not equal to [MDB_SUCCESS ](http://www.lmdb.tech/doc/group__errors.html#gad01fd0bf65d986197e7bf63024278ab6)(0) and throws an error
Regarding the usage of do while loops in macros, we referred this [link ](https://hownot2code.wordpress.com/2016/12/05/do-while-0-in-macros/) where they state it is best to use the do while loop in order to avoid getting any unexpected errors.

rrawther (approved):

PR #1012, opened 2022-12-19
ROCm 5.4.1 updates

kiritigowda:
* Docker Updates


docker/pytorch/mivisionx-with-pytorch.dockerfile
```diff
@@ -1,4 +1,3 @@
-
```
LakshmiKumar23, line 1: @kiritigowda Can you add the same workspace changes to this file as well?

PR #1036, opened 2023-02-03
rocAL - changes for classification training convergence

SundarRajan28:
Added changes required for classification training convergence

1. Added patch to compute resize dimensions using original image dimensions instead of decoded dimensions
2. Increased the max aspect ratio used for creating resize output buffer to 6 for accommodating the largest resize dimension in ImageNet validation dataset
3. Changed the DCT algorithm used in TurboJpeg decoder to accurate DCT algorithm which is the default algorithm used in DALI and PIL libraries
4. Added patch for using the correct rounding mode when calculating the crop dimensions
5. Added an else condition to choose another image in the same batch when the image fails decoding using TurboJpeg decoder
6. Changed the random aspect ratio range to the values used in torchvision's RandomResizedCrop function

### DIFF
--- a/rocAL/rocAL/include/pipeline/image.h
+++ b/rocAL/rocAL/include/pipeline/image.h
@@ -112,8 +112,8 @@ struct ImageInfo
     uint32_t * get_roi_height() const;
     const std::vector<uint32_t>& get_roi_width_vec() const;
     const std::vector<uint32_t>& get_roi_height_vec() const;
-    const std::vector<uint32_t>& get_original_roi_width_vec() const;
-    const std::vector<uint32_t>& get_original_roi_height_vec() const;
+    const std::vector<uint32_t>& get_original_width_vec() const;
+    const std::vector<uint32_t>& get_original_height_vec() const;
 
 private:
     Type _type = Type::UNKNOWN;//!< image type, whether is virtual image, created from handle or is a regular image
@@ -167,7 +167,7 @@ struct Image
 
     int create(vx_context context);
     void update_image_roi(const std::vector<uint32_t> &width, const std::vector<uint32_t> &height);
-    void update_image_original_roi(const std::vector<uint32_t> &original_width, const std::vector<uint32_t> &original_height);
+    void update_image_original_dims(const std::vector<uint32_t> &original_width, const std::vector<uint32_t> &original_height);
     void reset_image_roi() { _info.reallocate_image_roi_buffers(); }
     // create_from_handle() no internal memory allocation is done here since image's handle should be swapped with external buffers before usage
     int create_from_handle(vx_context context);
--- a/rocAL/rocAL/include/pipeline/image.h
+++ b/rocAL/rocAL/include/pipeline/image.h
@@ -112,8 +112,8 @@ struct ImageInfo
     uint32_t * get_roi_height() const;
     const std::vector<uint32_t>& get_roi_width_vec() const;
     const std::vector<uint32_t>& get_roi_height_vec() const;
-    const std::vector<uint32_t>& get_orig_roi_width_vec() const;
-    const std::vector<uint32_t>& get_orig_roi_height_vec() const;
+    const std::vector<uint32_t>& get_original_roi_width_vec() const;
+    const std::vector<uint32_t>& get_original_roi_height_vec() const;
 
 private:
     Type _type = Type::UNKNOWN;//!< image type, whether is virtual image, created from handle or is a regular image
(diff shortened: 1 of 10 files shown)


rocAL/rocAL/source/api/rocal_api_augmentation.cpp
```diff
@@ -29,7 +29,8 @@ THE SOFTWARE.
 #include "rocal_api.h"
 #include "image_source_evaluator.h"
 
-#define MAX_ASPECT_RATIO 3.0f
+// Calculated from the largest resize shorter dimension in imagenet validation dataset
+#define MAX_ASPECT_RATIO 6.0f
```
rrawther, line 33: how much accuracy change is achieved using this change?

rocAL/rocAL/include/pipeline/image.h
```diff
@@ -112,6 +112,9 @@ struct ImageInfo
     uint32_t * get_roi_height() const;
     const std::vector<uint32_t>& get_roi_width_vec() const;
     const std::vector<uint32_t>& get_roi_height_vec() const;
+    const std::vector<uint32_t>& get_original_roi_width_vec() const;
```
rrawther, line 115: please change the names to get_original_width_vec() && ..
roi is confusing

rocAL/rocAL/source/loaders/image/image_loader.cpp
```diff
@@ -288,7 +288,7 @@ ImageLoader::update_output_image()
     }
     _output_names = _output_decoded_img_info._image_names;
     _output_image->update_image_roi(_output_decoded_img_info._roi_width, _output_decoded_img_info._roi_height);
-
+    _output_image->update_image_original_roi(_output_decoded_img_info._original_width, _output_decoded_img_info._original_height);
```
rrawther, line 291: please remove "roi" from all the fucntions related to original dims

rocAL/rocAL/source/parameters/parameter_rocal_crop.cpp
```diff
@@ -53,8 +53,8 @@ void RocalCropParam::fill_crop_dims() {
             cropw_arr_val[img_idx] = (crop_w <= in_width[img_idx] && crop_w > 0) ? crop_w : in_width[img_idx];
             croph_arr_val[img_idx] = (crop_h <= in_height[img_idx] && crop_h > 0) ? crop_h : in_height[img_idx];
             if (_is_fixed_crop) {
-                x1_arr_val[img_idx] = static_cast<size_t>(_crop_anchor[0] * (in_width[img_idx] - cropw_arr_val[img_idx]));
-                y1_arr_val[img_idx] = static_cast<size_t>(_crop_anchor[1] * (in_height[img_idx] - croph_arr_val[img_idx]));
+                x1_arr_val[img_idx] = static_cast<size_t>(std::nearbyintf(_crop_anchor[0] * (in_width[img_idx] - cropw_arr_val[img_idx])));
```
rrawther, line 56: what is std::nearbyintf (I haven't heard of this)?
We need to use standard std:: round functions instead of internal functions within the library

rocAL/rocAL/source/pipeline/image.cpp
```diff
@@ -81,6 +81,16 @@ const std::vector<uint32_t>& ImageInfo::get_roi_height_vec() const
     return *_roi_height;
 }
 
+const std::vector<uint32_t>& ImageInfo::get_original_roi_width_vec() const
```
rrawther, line 84: please remove word from here "roi"

rocAL/rocAL/source/api/rocal_api_augmentation.cpp
SundarRajan28: Without this change, the rocAL classification trainings are not converging and we are achieving around 75.5x val top1 accuracy

rocAL/rocAL/source/parameters/parameter_rocal_crop.cpp
SundarRajan28: The round() functions round their argument to the nearest integer value in floating-point format, rounding halfway cases away from zero, regardless of the current rounding direction. 
The difference in the handling of these two variants of rounding "to nearest" is apparent when 0.5 is rounded to the next larger (in magnitude) integer (1) when using round(), but it rounds to an even integer (0) when nearbyint() is used. 
DALI uses round() function and it is producing different crop dimensions when compared to torchvision's crop dimensions. We made this change to ensure parity with torchvision's crop dimension calculation.

rocAL/rocAL/include/pipeline/image.h
SundarRajan28: Changed the function name in the latest commit

rocAL/rocAL/source/loaders/image/image_loader.cpp
SundarRajan28: Renamed all functions accordingly in the latest commit

rocAL/rocAL/source/pipeline/image.cpp
SundarRajan28: Changed the function name in the latest commit

rocAL/rocAL/source/pipeline/image.cpp
```diff
@@ -81,6 +81,16 @@ const std::vector<uint32_t>& ImageInfo::get_roi_height_vec() const
     return *_roi_height;
 }
 
+const std::vector<uint32_t>& ImageInfo::get_original_width_vec() const
+{
+    return *_original_roi_width;
```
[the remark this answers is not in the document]
SundarRajan28: Resolved in the latest commit

rrawther (approved):

rocAL/rocAL/source/parameters/parameter_rocal_crop.cpp
rrawther: The nearbyintf() functions assumes to set the floating point rounding mode before calling it (FE_TONEAREST). Otherwise it uses the default mode.

LakshmiKumar23:
@kiritigowda @rrawther Training converges. PR can be merged

LakshmiKumar23 (approved):

PR #1061, opened 2023-04-03
Documentation - Add configurations to build sphinx documentation on rtd for MIVisionX

samjwu:
Use rocm-docs-core template for documentation

Rationale:
 - https://github.com/RadeonOpenCompute/rocm-docs-core#purpose


kiritigowda:
@samjwu - I see that all the readme files in docs folder are replicates of .md files in the projects. This would cause out-of-sync documents. Is there a way we can copy these files on the fly when we are building?

saadrahim:
> @kiritigowda Yes, we can have them be copied over by Read the Docs whenever it needs to build documentation. The current issue is that the relative paths in the copied readme files are causing problems. I marked this as a draft partly for that reason the relative paths still need to be fixed.
> 
> I see two possibilities:
> 
> 1. have the readmes in the docs folder replace the others (ie: delete the old readmes to prevent syncing issues) or
> 2. change all relative paths into absolutes, and then have RTD copy them over as needed? (this means there would be no copy on the GitHub repo as RTD would be doing this on a different server)
> 
> Which is preferable?

Option 1 is preferable as it will standardize docs into one folder. That is also the intent of the pitchfork methodology for repository organization.

kiritigowda:
> > @kiritigowda Yes, we can have them be copied over by Read the Docs whenever it needs to build documentation. The current issue is that the relative paths in the copied readme files are causing problems. I marked this as a draft partly for that reason the relative paths still need to be fixed.
> > I see two possibilities:
> > 
> > 1. have the readmes in the docs folder replace the others (ie: delete the old readmes to prevent syncing issues) or
> > 2. change all relative paths into absolutes, and then have RTD copy them over as needed? (this means there would be no copy on the GitHub repo as RTD would be doing this on a different server)
> > 
> > Which is preferable?
> 
> Option 1 is preferable as it will standardize docs into one folder. That is also the intent of the pitchfork methodology for repository organization.

@saadrahim Option 1 will change how most of our users access documents on GitHub. The current structure keeps the docs & modules together. Is there a way we can keep the current structure and let the document build script handle everything else? We can discuss in details over a call?

samjwu:
Preview: https://advanced-micro-devices-mivisionx--1061.com.readthedocs.build/en/1061/README.html

kept the same file structure on GitHub for existing users
- `conf.py` copies over the READMEs for Read the Docs to serve

updated to appear correct on both GitHub and Read the Docs
- for images, updated the relative paths to absolute paths 
- for README links, updated relative paths to include name of README file so that

***NOTE***: preview will not have images yet since they were moved to /data/ directory for standardization across documentation. images will appear when this is merged and the images are in the correct location

saadrahim:
The toc.yml needs to be built for this project. Right now the left nav links just contain changelog.

samjwu:
> The toc.yml needs to be built for this project. Right now the left nav links just contain changelog.

[update toc to match main README](https://github.com/GPUOpen-ProfessionalCompute-Libraries/MIVisionX/pull/1061/commits/a71e427b31d2fa70e76e943a2bfaea203ee05e99) adds all the separate doc pages to the left TOC/navbar roughly in the order they appear in the main README

rrawther (approved):
@kiritigowda to verify

LakshmiKumar23 (approved):

PR #1063, opened 2023-04-10
remove shuffle_time from rocal timing

rrawther:
Shuffle time is not very significant. It was added for debugging and removing since it won't be used


LakshmiKumar23 (approved):
@kiritigowda can be merged

Issue #138, opened 2019-06-13
Developer - [OpenVX] AGO_DUMP_GRAPH - dumps wrong format for GDF


LakshmiKumar23:
Wrong format of GDF is dumped on setting the flag. Example: tensor-virtual needs to be replaced with virtual-tensor.

kiritigowda:
reopen this issue if the feature is required.

PR #1064, opened 2023-04-12
rocAL - MaskRCNN training support

SundarRajan28:
- Metadata changes for supporting polygon masks in rocAL
- Changing bbox datatype to double and adding the necessary rocAL and pybind changes
- rocAL pybind changes for RMN augmentation and reader changes
- Adding a new example python pipeline which can be run to check the rocAL segmentation pipeline


rocAL/rocAL/include/api/rocal_api_info.h
```diff
@@ -33,6 +33,18 @@ extern "C" int ROCAL_API_CALL rocalGetOutputWidth(RocalContext rocal_context);
 /// \return The height of the ROCAL's output image in pixels. It includes all images in the batch.
 extern "C" int ROCAL_API_CALL rocalGetOutputHeight(RocalContext rocal_context);
 
+///
+/// \param rocal_context Rocal context
+/// \param buf user buffer provided to be filled with output image width for images in the output batch.
+/// \return The width of the ROCAL's output image in pixels
+extern "C" void ROCAL_API_CALL rocalGetOutputResizeWidth(RocalContext rocal_context, unsigned int* buf);
```
rrawther, line 40: please combine both apis into single one for returning ROI

rocAL/rocAL/include/api/rocal_api_meta_data.h
```diff
@@ -131,11 +131,23 @@ extern "C" void ROCAL_API_CALL rocalGetImageLabels(RocalContext rocal_context, v
...
+/// \param rocal_context
+/// \param buf the imageIdx in the output batch
+/// \return The size of the buffer needs to be provided by user to get mask box info associated with image_idx in the output batch.
+extern "C" unsigned ROCAL_API_CALL rocalGetMaskCount(RocalContext rocal_context, int* buf );
+
+///
+/// \param rocal_context
+/// \param bufcount The user's buffer that will be filled with polygon size for the mask info
+/// \param buf The user's buffer that will be filled with mask info for the images in the output batch. It needs to be of size returned by a call to the rocalGetMaskCount
+extern "C" void ROCAL_API_CALL rocalGetMaskCoordinates(RocalContext rocal_context, int* bufcount, float* buf);
```
rrawther, line 144: I think mask_count can be extracted from this api. In that case do you need the api on #138

rocAL/rocAL/include/api/rocal_api_meta_data.h
```diff
@@ -176,7 +188,7 @@ extern "C" void ROCAL_API_CALL rocalBoxEncoder(RocalContext p_context, std::vect
 
 /// \param boxes_buf  user's buffer that will be filled with encoded bounding boxes . Its needs to be at least of size batch_size.
 /// \param labels_buf  user's buffer that will be filled with encoded labels . Its needs to be at least of size batch_size.
-extern "C" void ROCAL_API_CALL rocalCopyEncodedBoxesAndLables(RocalContext p_context, float* boxes_buf, int* labels_buf);
+extern "C" void ROCAL_API_CALL rocalCopyEncodedBoxesAndLables(RocalContext p_context, double* boxes_buf, int* labels_buf);
```
rrawther, line 191: Does the training uses encoded boxes and labels in double? Otherwise we should return in float. We can do the internal calc in double, but keep the data at the end in float

rocAL/rocAL/include/meta_data/meta_data.h
```diff
@@ -160,19 +187,28 @@ struct MetaDataBatch
...
     virtual std::shared_ptr<MetaDataBatch> clone()  = 0;
+    void set_metadata_type(MetaDataType metadata_type) { _type = metadata_type; }
+    MetaDataType metadata_type() { return _type; }
     std::vector<int>& get_label_batch() { return _label_id; }
     std::vector<BoundingBoxCords>& get_bb_cords_batch() { return _bb_cords; }
     std::vector<BoundingBoxCords_xcycwh>& get_bb_cords_batch_xcycxwh() { return _bb_cords_xcycwh; }
     std::vector<BoundingBoxLabels>& get_bb_labels_batch() { return _bb_label_ids; }
     ImgSizes & get_img_sizes_batch() { return _img_sizes; }
     JointsDataBatch & get_joints_data_batch() { return _joints_data; }
+    std::vector<MaskCords>& get_mask_cords_batch() { return _mask_cords; }
```
rrawther, line 198: is this different than mask_vertices?

rocAL/rocAL/source/api/rocal_api_augmentation.cpp
```diff
@@ -601,13 +601,9 @@ rocalResizeMirrorNormalize(
...
 
         std::shared_ptr<ResizeMirrorNormalizeNode> rmn_node =  context->master_graph->add_node<ResizeMirrorNormalizeNode>({input}, {output});
-        // RPP doesn't support returning float buffers so passing 0 and 1 as mean and std and doing normalization in rocAL
-        // TODO: To be removed with rocAL Tensor support
-        // rmn_node->init(0, 1, mirror);
         rmn_node->init(mean, std_dev, mirror);
-        // TODO: Uncomment the below lines once RMN meta node is added to ToT
-        // if (context->master_graph->meta_data_graph())
-        //     context->master_graph->meta_add_node<ResizeMirrorNormalizeMetaNode,ResizeMirrorNormalizeNode>(rmn_node);
+        if (context->master_graph->meta_data_graph())
```
rrawther, line 605: where is the init called here?

rocAL/rocAL/source/api/rocal_api_meta_data.cpp
```diff
@@ -77,12 +77,13 @@ ROCAL_API_CALL rocalCreateVideoLabelReader(RocalContext p_context, const char* s
...
 RocalMetaData
-ROCAL_API_CALL rocalCreateCOCOReader(RocalContext p_context, const char* source_path, bool is_output){
+ROCAL_API_CALL rocalCreateCOCOReader(RocalContext p_context, const char* source_path, bool is_output, const bool mask) {
     if (!p_context)
         THROW("Invalid rocal context passed to rocalCreateCOCOReader")
     auto context = static_cast<Context*>(p_context);
-
-    return context->master_graph->create_coco_meta_data_reader(source_path, is_output, MetaDataReaderType::COCO_META_DATA_READER,  MetaDataType::BoundingBox);
+    if(mask)
+        return context->master_graph->create_coco_meta_data_reader(source_path, is_output, MetaDataReaderType::COCO_META_DATA_READER, MetaDataType::PolygonMask);
```
rrawther, line 85: can't we combine both mask and bounding bix reading in one single function. Going through the dataset twice is very inefficient

rocAL/rocAL/source/api/rocal_api_meta_data.cpp
```diff
@@ -368,6 +369,65 @@ ROCAL_API_CALL rocalGetBoundingBoxCords(RocalContext p_context, float* buf)
     }
 }
 
+unsigned
+ROCAL_API_CALL rocalGetMaskCount(RocalContext p_context, int* buf)
+{
+    if (p_context == nullptr)
+        THROW("Invalid rocal context passed to rocalGetMaskCount")
```
rrawther, line 376: probably need to return from here instead of throwing an exception here

rocAL/rocAL/source/meta_data/meta_data_graph_factory.cpp
```diff
@@ -38,6 +38,10 @@ std::shared_ptr<MetaDataGraph> create_meta_data_graph(const MetaDataConfig& conf
         {
             return std::make_shared<BoundingBoxGraph>();
         }
+        case MetaDataType::PolygonMask:
```
rrawther, line 41: can you please combine 3 cases into single switch branch since they are returning the same data

rocAL/rocAL/source/meta_data/meta_node_resize_mirror_normalize.cpp
```diff
@@ -0,0 +1,100 @@
...
+        auto bb_count = input_meta_data->get_bb_labels_batch()[i].size();
+        BoundingBoxCords coords_buf;
+        BoundingBoxLabels labels_buf;
+        coords_buf.resize(bb_count);
+        labels_buf.resize(bb_count);
+        memcpy(labels_buf.data(), input_meta_data->get_bb_labels_batch()[i].data(), sizeof(int) * bb_count);
+        memcpy((void *)coords_buf.data(), input_meta_data->get_bb_cords_batch()[i].data(), input_meta_data->get_bb_cords_batch()[i].size() * sizeof(BoundingBoxCord));
+        BoundingBoxCords bb_coords;
+        BoundingBoxLabels bb_labels;
+        if (input_meta_data->metadata_type() == MetaDataType::PolygonMask)
```
rrawther, line 65: doing lot of computations in update_parameters() is not recommended. Need to time this code and see adding openMP pragma can help improve performance.

rocAL/rocAL_pybind/amd/rocal/plugin/pytorch.py
```diff
@@ -150,7 +150,7 @@ def __next__(self):
             self.labels = np.zeros(self.count_batch, dtype="int32")
             self.loader.GetBBLabels(self.labels)
             # 1D bboxes array in a batch
-            self.bboxes = np.zeros((self.count_batch*4), dtype="float32")
+            self.bboxes = np.zeros((self.count_batch*4), dtype="float64")
```
rrawther, line 153: do we need to store the boxes in float64

rocAL/rocAL_pybind/examples/rocAL_api_coco_pipeline_segmentation.py
```diff
@@ -0,0 +1,555 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+import torch
+import random
```
rrawther, line 5: removed unused

rocAL/rocAL/source/api/rocal_api_meta_data.cpp
rrawther, line 85: can't we read the polygons mask and bounding boxes from the same api. Going through the dataset twice is inefficient

rocAL/rocAL_pybind/examples/rocAL_api_coco_pipeline_segmentation.py
```diff
@@ -0,0 +1,555 @@
...
+        self.w = self.loader.getOutputWidth()
+        self.h = self.loader.getOutputHeight()
+        self.n = self.loader.getOutputImageCount()
+        self.rim = self.loader.getRemainingImages()
+        self.display = display
+        print("____________REMAINING IMAGES____________:", self.rim)
+        color_format = self.loader.getOutputColorFormat()
+        self.p = (1 if color_format is types.GRAY else 3)
+        if tensor_layout == types.NCHW:
+            if self.device == "cpu":
```
rrawther, line 329: please combine both cpu and gpu cases into single line 
device = None if device==cpu else torch_gpu_device

rocAL/rocAL_pybind/examples/rocAL_api_coco_pipeline_segmentation.py
```diff
@@ -0,0 +1,555 @@
...
+        self.lis = []  # Empty list for bboxes
+        self.lis_lab = []  # Empty list of labels
+
+        self.loader.copyToTensor(
+            self.out, self.multiplier, self.offset, self.reverse_channels, self.tensor_format, self.tensor_dtype)
+
+
+        self.img_names_length = np.empty(self.bs, dtype="int32")
+        self.img_names_size = self.loader.GetImageNameLen(self.img_names_length)
+        # print("Image name length:", self.img_names_size)
```
rrawther, line 377: please remove all commented debug statements

rocAL/rocAL/source/api/rocal_api_meta_data.cpp
rrawther: It is also not ok to THROW exception in an API without a try catch loop. You can return a invalid return value if the context is not valid instead of throw()

LakshmiKumar23:
@SundarRajan28 can you fix the merge conflicts on this PR?

rrawther:
@SundarRajan28 : Can you please merge the changes for working HIP backend for mask-rcnn? I think upstream changes needs to be merged as well

rrawther (changes requested):
@SundarRajan28 Looks like the tot changes are not merged

rocAL/rocAL/source/pipeline/master_graph.cpp
```diff
@@ -1070,6 +1090,16 @@ void MasterGraph::output_routine()
                 else
                     full_batch_meta_data = _augmented_meta_data->clone();
             }
+            // get roi width and height of output image
```
rrawther, line 1093: I don't understand the logic here. Why are we using another array for resize_width and resize_height? _output_image_info assumed to be constant for all the images and reinitializing every time here won't change anything for resize_width and height. In this case it is better to use output_image_info directly in the resize_width and resize_height api calls

rrawther (changes requested):
please address review comments

rocAL/rocAL/include/api/rocal_api_meta_data.h
SundarRajan28: rocalGetMaskCount - returns the number of polygons for each bbox annotation for an image
rocalGetMaskCoordinates - returns the number of mask values in each polygon for each bbox annotation for an image

rocAL/rocAL/include/api/rocal_api_meta_data.h
SundarRajan28: The mask training doesn't use this API or box encoder. Changed the dtype of the API parameter since the bboxes dtype is changed to double in metadata.h

rocAL/rocAL/include/meta_data/meta_data.h
SundarRajan28: The polygon and vertices names are used to represent the counts only. Mask data values are represented as mask_coords

rocAL/rocAL/source/api/rocal_api_augmentation.cpp
SundarRajan28: meta_data_graph is created in this function in master_graph.cpp - create_coco_meta_data_reader

rocAL/rocAL/source/api/rocal_api_meta_data.cpp
SundarRajan28: 1. The coco_meta_data_reader reads both polygon masks and bboxes in a single read
2. The THROW exception was already present in this function and all the other reader creation functions

rocAL/rocAL/source/meta_data/meta_node_resize_mirror_normalize.cpp
SundarRajan28: Will add timers and check if this is a bottleneck

rocAL/rocAL_pybind/amd/rocal/plugin/pytorch.py
SundarRajan28: I'm returning the bboxes as double so needs float64 to store the boxes. I'll implement the change we discussed about return boxes as float32

rocAL/rocAL/source/api/rocal_api_meta_data.cpp
SundarRajan28: The coco_meta_data_reader reads both polygon masks and bboxes in a single read. Depending on MetaDataType passed to  create_coco_meta_data_reader, it either reads boxes only or boxes and polygon masks

rocAL/rocAL/source/pipeline/master_graph.cpp
SundarRajan28: The reasoning behind this implementation was that the _resize_width and _resize_height acts as a queue which inserts the ROI info of each batch in a new vector at the front. The output_resize_width function in the same file pops the last ROI vector and returns it to the vector. Due to prefetch_queue_depth, we couldn't directly fetch the ROI info from ouptut_image_info and return it to the user. I tried that as part of the experiments to fix the HIP backend issue and it gave wrong values.
The fix for the HIP backend mask training removes this vector updation and the respective ROI API functions. Instead I'm adding the ROI as an internal ImgSize struct variable of MetaDataBatch and updating it in meta_node for RMN. I'll be pushing the new ROI implementation soon after the internal review

PR #1096, opened 2023-06-02
hwdec fix bug which was causing crash in MI300

rrawther:
Fixes the bug which was causing crash in MI300 when using hwdec


LakshmiKumar23 (approved):
@rrawther @kiritigowda fixes the issue noticed on image augmentation app on MI300. Can be merged

kiritigowda (approved):

PR #1158, opened 2023-08-03
CTest - Updates & Additional Tests

kiritigowda:
* Fix old format
* Add additional tests


kiritigowda:
@hansely & @paveltc to run tests

```
cmake ../MIVisionX
make -j8
sudo make install
```

### Test with basic info
```
make test
```

### Test with detailed info
```
make test ARGS="-VV"
```

PR #1176, opened 2023-08-26
VX_NN FP16 bug fix

hansely:
-fixed a bug where fp16 conversion doesn't work when there are mixed weight types in the graph
-added fp16 support for argmax layer
-fixed stride value for tensor matrix multiplication

### DIFF
--- a/amd_openvx_extensions/amd_nn/src/tensor_matrix_multiply.cpp
+++ b/amd_openvx_extensions/amd_nn/src/tensor_matrix_multiply.cpp
@@ -200,7 +200,7 @@ static vx_status VX_CALLBACK initialize(vx_node node, const vx_reference *parame
     data->tA = params.transpose_input1 ? true : false;
     data->tB = params.transpose_input2 ? true : false;
     data->tI = params.transpose_input3 ? true : false;
-   if (input1_dims[2]&input1_dims[3]) {
+    if (input1_dims[2]&input1_dims[3]) {
         data->k = input1_dims[params.transpose_input1 ? 1 : 0];
         data->m = input1_dims[params.transpose_input1 ? 0 : 1];
     }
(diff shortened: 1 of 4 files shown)


amd_openvx_extensions/amd_nn/src/tensor_matrix_multiply.cpp
```diff
@@ -200,7 +200,7 @@ static vx_status VX_CALLBACK initialize(vx_node node, const vx_reference *parame
     data->tA = params.transpose_input1 ? true : false;
     data->tB = params.transpose_input2 ? true : false;
     data->tI = params.transpose_input3 ? true : false;
-   if (input1_dims[2]&input1_dims[3]) {
+    if (input1_dims[2]&input1_dims[3]) {
```
asalmanp, line 203: formatting: add a space before and after "&"

asalmanp (changes requested):

paveltc:
@hansely Do you have a unit test for this?

hansely:
> @hansely Do you have a unit test for this?

For each layers(argmax & tensor matmul), you can run the corresponding unit tests available at MIVisionX-tests repo with fp16 inputs.
For testing the fp16 conversion flow, mivisionx-onnx-run.py script with fp16=1 setting should be sufficient.

paveltc:
@hansely The tests in the openvx-neural-net-test-suite were made by you and Lakshmi with specific inputs and golden outputs to check if the layers were given the correct answers. They were compared against the outputs of Caffe.
Some tests are in GDF format and some are C Code unit tests because runvx does not support certain parameters for the layers. MatMul is a layer that has a C Code test. 
Can you please take a look and see if you can make some unit tests with golden outputs for these layers?

asalmanp (approved):

PR #1177, opened 2023-08-28
Model Compiler - Update & Bug Fixes

kiritigowda:
## Model Compiler 
* include dirs updated
* rocm path added
* Fix bugs

### DIFF
--- a/model_compiler/python/nnir_to_clib.py
+++ b/model_compiler/python/nnir_to_clib.py
@@ -161,9 +161,9 @@ if (OPENVX_BACKEND_OPENCL_FOUND)
 endif()
 
 find_package(OpenCV QUIET)
-include_directories (/opt/rocm/include/mivisionx)
+include_directories (${ROCM_PATH}/include ${ROCM_PATH}/include/mivisionx)
 include_directories (${PROJECT_SOURCE_DIR}/lib)
-link_directories    (/opt/rocm/lib)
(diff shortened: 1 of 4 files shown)


model_compiler/python/nnir_to_clib.py
```diff
@@ -298,8 +298,8 @@ def generateCMakeExtras(graph, outputFolder):
 endif()
 
 find_package(OpenCV QUIET)
-include_directories (/opt/rocm/include/miviisionx ../)
-link_directories    (/opt/rocm/lib)
+include_directories (${ROCM_PATH}/include ${ROCM_PATH}/include/mivisionx ../)
```
kiritigowda, line 301: @hansely can you confirm why we have `../` as one of the dirs? is this typo or can we fix this with a full link?

hansely: @kiritigowda sorry that is a typo. Would you be able to remove it in this PR?

paveltc:
@kiritigowda What are the relevant tests for this PR?

kiritigowda:
@paveltc - This PR would need to run NN tests and mv_compile tests to validate. @hansely add any tests you think we need apart for those?

hansely:
@paveltc @kiritigowda The only related app to nnir_to_clib.py script is mv_compile so far. Therefore, we won't need any extra unit test for this PR.

hansely (approved):

PR #1224, opened 2023-11-04
MIVisionX ASAN - Exclude library_tests from ASAN pkg

arvindcheru:
Changes Proposed:
- Exclude test files from MIVisionX ASAN package.
- Fix for SWDEV-430827


arvindcheru:
> @arvindcheru I guess, we need to keep it 1st in "develop" branch then from there it needs to moved to "master"
> 
> Also, I have one query. I see, we are disabling binary file for ASAN? When we have test package enabled, we need those binary files, correct?
> 
> Feature ticket for asan instrument test package , where this binary is needed, https://ontrack-internal.amd.com/browse/SWDEV-429801

kiritigowda:
@arvindcheru - some observation in the change. The only libraries and executables which MIVisionX installs is listed below. The apps and utilities are not built with TOT CMakeLists. Changes for ASAN will never propagate to those CMake.

## Libraries
```
ibopenvx.so
libvxu.so
libvx_opencv.so
libvx_amd_media.so
libvx_amd_custom.so
libvx_nn.so
libvx_amd_migraphx.so
libvx_rpp.so
librocal.so
```

##  Executables
```
runvx
runcl
loom_shell
mv_compile
```

arvindcheru:
> Changes Proposed:
> 
> * Exclude bin and test files from MIVisionX ASAN package.



> @arvindcheru - some observation in the change. The only libraries and executables which MIVisionX HIP installs is listed below. The apps and utilities are not built with TOT CMakeLists. Changes for ASAN will never propagate to those CMake.
> 
> ## Libraries
> ```
> ibopenvx.so
> libvxu.so
> libvx_opencv.so
> libvx_amd_media.so
> libvx_amd_custom.so
> libvx_nn.so
> libvx_amd_migraphx.so
> libvx_rpp.so
> librocal.so
> ```
> 
> ## Executables
> ```
> runvx
> mv_compile
> ```

Updated the change as discussed considering the above review comments

kiritigowda (approved):

PR #1244, opened 2023-12-14
Fixing linkage error for std experimental filesystem

SundarRajan28:
Building MIVisionX with std::experimental::filesystem needs stdc++fs linkage in cmake. Without this, rocAL builds without errors but unit tests build fails.
```
/usr/bin/ld: /opt/rocm/lib/librocal.so: undefined reference to `std::experimental::filesystem::v1::__cxx11::path::_M_find_extension() const'
/usr/bin/ld: /opt/rocm/lib/librocal.so: undefined reference to `std::experimental::filesystem::v1::status(std::experimental::filesystem::v1::__cxx11::path const&)'
/usr/bin/ld: /opt/rocm/lib/librocal.so: undefined reference to `std::experimental::filesystem::v1::__cxx11::path::_M_split_cmpts()'
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/rocAL_unittests.dir/build.make:99: rocAL_unittests] Error 1
make[1]: *** [CMakeFiles/Makefile2:76: CMakeFiles/rocAL_unittests.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
```


kiritigowda (approved):

kiritigowda:
@SundarRajan28 - this is only required for experimental FS. @hansely can you confirm?

kiritigowda:
@SundarRajan28 we may want to add stdc++fs if experimental fs found and not other versions?

SundarRajan28:
@kiritigowda As shared in this CMake [issue](https://gitlab.kitware.com/cmake/cmake/-/issues/17834), the std::experimental::filesystem will be added to libstdc++.so eventually. Until then it remains as a static library and needs to be linked seperately. So if std::filesystem is detected, there's no need for linking this static library

PR #1248, opened 2023-12-15
Std filesystem cmake fix

SundarRajan28:
Even for newer gcc and clang versions, the -std=gnu++17 argument must be added for supporting std::filesystem features while compiling. Due to the missed argument, even gcc versions > 9 with std::filesystem added in libstdc++ fails the compilation test and std::experimental::filesystem is used instead.

### DIFF
--- a/rocAL/rocAL/CMakeLists.txt
+++ b/rocAL/rocAL/CMakeLists.txt
@@ -42,7 +42,7 @@ find_package(RapidJSON QUIET)
 
 # Check for filesystem support
 include(CheckCXXSourceCompiles)
-set(CMAKE_REQUIRED_FLAGS "${CMAKE_CXX_FLAGS}")
+set(CMAKE_REQUIRED_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++17")
 set(CMAKE_REQUIRED_INCLUDES "")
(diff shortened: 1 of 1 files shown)


rocAL/rocAL/CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ find_package(RapidJSON QUIET)
 
 # Check for filesystem support
 include(CheckCXXSourceCompiles)
-set(CMAKE_REQUIRED_FLAGS "${CMAKE_CXX_FLAGS}")
+set(CMAKE_REQUIRED_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++17")
```
kiritigowda, line 45: https://github.com/ROCm/MIVisionX/blob/ae9c9ece492859bd50bf66c88e6cf5697630fd79/rocAL/rocAL/CMakeLists.txt#L271

adds the CXX Flags

SundarRajan28: [https://github.com/ROCm/MIVisionX/blob/master/rocAL/rocAL/CMakeLists.txt#L47](url)
This line runs the std filesystem checks before the CXX flags are set so unless they are updated before this check, std filesystem check always fails

kiritigowda (approved):

PR #1275, opened 2024-01-26
Develop - Merge to master

kiritigowda:
* Lint Fixes


rocAL/rocAL_pybind/amd/rocal/fn.py
```diff
@@ -19,11 +19,6 @@
 # THE SOFTWARE.
 
 
-from amd.rocal import readers
```
kiritigowda, line 22: @LakshmiKumar23 - can you verify this change for rocAL?

LakshmiKumar23: @kiritigowda These are required. Please make sure this file is reverted

PR #1287, opened 2024-02-16
CMakeLists.txt update to remove unnecessary variable

hansely:
Remove unnecessary variable : MIVisionX_LIBRARIES_DIRS

### DIFF
--- a/cmake/FindMIVisionX.cmake
+++ b/cmake/FindMIVisionX.cmake
@@ -64,19 +64,6 @@ find_library(VXRPP_LIBRARIES
 )
 mark_as_advanced(VXRPP_LIBRARIES)
 
-find_path(MIVisionX_LIBRARIES_DIRS
-    NAMES libopenvx${SHARED_LIB_TYPE}
-    HINTS
-    $ENV{ROCM_PATH}/lib
-    $ENV{ROCM_PATH}/lib64
-    $ENV{MIVisionX_PATH}/lib
-    PATHS
-    ${MIVisionX_PATH}/lib
-    /usr/lib
-    ${ROCM_PATH}/lib
-)
-mark_as_advanced(MIVisionX_LIBRARIES_DIRS)
-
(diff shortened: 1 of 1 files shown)


cmake/FindMIVisionX.cmake
```diff
@@ -64,19 +64,6 @@ find_library(VXRPP_LIBRARIES
...
-    HINTS
-    $ENV{ROCM_PATH}/lib
-    $ENV{ROCM_PATH}/lib64
-    $ENV{MIVisionX_PATH}/lib
-    PATHS
-    ${MIVisionX_PATH}/lib
-    /usr/lib
-    ${ROCM_PATH}/lib
-)
-mark_as_advanced(MIVisionX_LIBRARIES_DIRS)
```
kiritigowda, line 78: Variable used when custom installed by user.

cmake/FindMIVisionX.cmake
```diff
@@ -64,19 +64,6 @@ find_library(VXRPP_LIBRARIES
 )
 mark_as_advanced(VXRPP_LIBRARIES)
 
-find_path(MIVisionX_LIBRARIES_DIRS
```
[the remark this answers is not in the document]
hansely: For inference server docker(ubuntu 20.04), using find_path() to find .so file (libopenvx.so in this case) was failing. 
Also, the FindMIVisionX.cmake file already uses find_library() to find libopenvx.so. 
Why do we need two different variables : OPENVX_LIBRARIES & MIVisionX_LIBRARIES_DIRS that points to the same .so file?

cmake/FindMIVisionX.cmake
hansely: If that's the case, can we use OPENVX_LIBRARIES variable instead? It points to the same libopenvx.so file.

kiritigowda (approved):

PR #1356, opened 2024-05-20
Fixing HIP CMake issue

SundarRajan28:
Setting AMDGPU_TARGETS before find_package(HIP) call - fixes HIP issues


CMakeLists.txt
```diff
@@ -121,6 +121,8 @@ message("-- ${BoldBlue}MIVisionX Build Type -- ${CMAKE_BUILD_TYPE}${ColourReset}
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 list(APPEND CMAKE_PREFIX_PATH ${ROCM_PATH} ${ROCM_PATH}/hip)
+set(DEFAULT_AMDGPU_TARGETS "gfx908;gfx90a;gfx940;gfx941;gfx942;gfx1030;gfx1031;gfx1032;gfx1100;gfx1101;gfx1102")
```
kiritigowda, line 124: Top level cmakelist cannot have GPU targets, we need to add this to all projects that have HIP compile.

SundarRajan28: Then can we remove the find_package(HIP) call in the Top level CMake and use it wherever we set the GPU targets?

LakshmiKumar23: Yes Sundar, please use find_package(hip) wherever we have the targets built already

SundarRajan28:
@kiritigowda @LakshmiKumar23 Removed find_package(HIP) from top level CMake

kiritigowda (approved):

PR #1394, opened 2024-07-15
Ubuntu 24.04 - Support Updates

kiritigowda:
* Setup updates
* Test updates


kiritigowda:
@LakshmiKumar23 -- we use the import futures in our NN test and with python 3.12 we might not be able support that. Any thoughts on removing this?

```
File "/opt/venv/lib/python3.12/site-packages/future/standard_library/__init__.py", line 65, in <module>
12:     import imp
12: ModuleNotFoundError: No module named 'imp'
12: Traceback (most recent call last):
12:   File "/opt/rocm/libexec/mivisionx/model_compiler/python/nnir_update.py", line 25, in <module>
12:     from future import standard_library
12:   File "/opt/venv/lib/python3.12/site-packages/future/standard_library/__init__.py", line 65, in <module>
12:     import imp
```

PR #1416, opened 2024-08-23
Fixing Resample build issue

SundarRajan28:
Fixes the build issue with ToT rpp due to repeated definition of windowed_sinc function for Resample kernel


SundarRajan28:
@kiritigowda This PR will fail when rpp is installed using package as the windowed_sinc function is only added in rppdefs.h two days ago. The build error is only seen with ToT rpp

rrawther (approved):

danielsu-amd:
Sorry about that, not sure how I closed this PR. I found there was an error in our Azure CI that caused it to pull old builds of rpp, this should be fixed now.

PR #1436, opened 2024-10-04
NN Dependencies - Bump future package from 0.18.2 to 1.0.0

hansely:
Since Python 3.12, the `imp` module has been deprecated, which caused the following error when running NN tests:

`
ModuleNotFoundError: No module named 'imp'
`

This PR upgrades the setup script and python test scripts to install the future package version 1.0.0, which is compatible with the latest Python versions.

### DIFF
--- a/MIVisionX-setup.py
+++ b/MIVisionX-setup.py
@@ -282,7 +282,7 @@ if "VERSION_ID=24" in os_info_data:
     pipONNXVersion = "onnx==1.16.0"
     pipProtoVersion= "protobuf==3.20.2"
 pip3InferencePackagesUbuntu = [
-    'future==0.18.2',
+    'future==1.0.0',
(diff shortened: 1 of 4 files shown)


kiritigowda (approved):

kiritigowda:
@hansely - merge pending CI results

MIVisionX-setup.py
```diff
@@ -302,7 +302,7 @@ def ERROR_CHECK(call):
     pipNNEFversion = "protobuf==3.19.5" # TBD: NO NNEF Package for SLES
 
 pip3InferencePackagesRPM = [
-    'future==0.18.2',
```
kiritigowda, line 305: @hansely should be keep the RPM version unchanged as I am not sure if 1.0.0 is available on RHEL 8

PR #1483, opened 2025-02-03
gfx updates

LakshmiKumar23:
Updating gfx list


kiritigowda (approved):

LakshmiKumar23:
@kiritigowda CI is failing with MIOpen errors: can you check?
```./anntest: error while loading shared libraries: libMIOpen.so.1: cannot open shared object file: No such file or directory```

PR #1491, opened 2025-02-28
Adding Audio CMake fix for RPP v2.0

SundarRajan28:


amd_openvx_extensions/amd_rpp/CMakeLists.txt
```diff
@@ -211,7 +211,8 @@ if(EXISTS ${RPP_INCLUDE_DIR}/rpp_audio_augmentations_support.h)
 endif()
 
 # Audio features for VX_RPP
-if(${RPP_VERSION_MAJOR} VERSION_GREATER_EQUAL "1" AND ${RPP_VERSION_MINOR} VERSION_GREATER_EQUAL "8" AND ${RPP_VERSION_PATCH} VERSION_GREATER_EQUAL "0" AND RPP_AUDIO_AUGMENTATIONS_SUPPORT_FOUND)
```
kiritigowda, line 214: @SundarRajan28 -- `VERSION_GREATER_EQUAL` - should work - why add `equal` & `greater` separate?

[cmake ref](https://cmake.org/cmake/help/latest/variable/CMAKE_VERSION.html)

SundarRajan28: VERSION_GREATER_EQUAL fails for RPP v2.0.0 since its checking if each individual version (major, minor and patch) >= 1.8.0
I added a new condition to check if the major version > 1 and if so, it doesn't check for the minor and patch version and enables audio support

r-abishek: Yes in the current ToT version, the middle condition will fail for RPP 2.0.0 since it'll look for "minor version >= 8".

So now, given RPP_AUDIO_AUGMENTATIONS_SUPPORT_FOUND is true, basically the checks are to enable under these two version conditions only:
- Either major > 1 OR
- If major = 1 and minor >= 8 and patch >= 0

Looks good.

r-abishek (approved):

PR #1496, opened 2025-03-14
VX_RPP - Transpose and normalize augmentation support

SundarRajan28:
- Adding vx_rpp extension for transpose node
- Adding changes to normalize node for supporting NHWC/NCHW layout


amd_openvx_extensions/amd_rpp/source/tensor/Normalize.cpp
```diff
@@ -77,10 +88,40 @@ static vx_status VX_CALLBACK refreshNormalize(vx_node node, const vx_reference *
             dst_roi[i].xywhROI.roiHeight = src_roi[i].xywhROI.roiHeight;
         }
         data->pSrcRoi = static_cast<unsigned *>(data->pSrcDims);
+    } else if (numDims == 4) {
+        RpptROI *src_roi = reinterpret_cast<RpptROI *>(roi_tensor_ptr);
+        RpptROI *dst_roi = reinterpret_cast<RpptROI *>(roi_tensor_ptr_dst);
+        for (unsigned i = 0; i < data->inputTensorDims[0]; i++) {
+            // rocAL ROI for image formats is stored in XYWH format. Transpose kernel needs ROI for all dims so adding the channel ROI here
+            unsigned index = i * nDim * 2;
+            if (data->inputLayout == vxTensorLayout::VX_NHWC) {
```
rrawther, line 97: I am confused about this data conversion? what is the difference between nDims and numDims?
Also NHWC means channel will be 0 index. So data->pSrcDims[index + 0] represents channel roi, isn't it?

amd_openvx_extensions/amd_rpp/source/tensor/Transpose.cpp
```diff
@@ -0,0 +1,299 @@
...
+        STATUS_ERROR_CHECK(vxQueryTensor((vx_tensor)parameters[2], VX_TENSOR_BUFFER_HOST, &data->pDst, sizeof(data->pDst)));
+        if (!data->pPerm) data->pPerm = new unsigned[nDim];
+        if (!data->pSrcRoi && (numDims == 4)) {
+            data->pSrcRoi = new unsigned[data->inputTensorDims[0] * nDim * 2];
+        }
+        STATUS_ERROR_CHECK(vxCopyArrayRange((vx_array)parameters[3], 0, nDim, sizeof(unsigned), data->pPerm, VX_READ_ONLY, VX_MEMORY_TYPE_HOST));
+    }    
+    if (numDims == 4) {
+        RpptROI *src_roi = reinterpret_cast<RpptROI *>(roi_tensor_ptr);
+        for (unsigned i = 0; i < data->inputTensorDims[0]; i++) {
```
rrawther, line 79: same as previous comment

rrawther (changes requested):
Added a comment

amd_openvx_extensions/amd_rpp/source/tensor/Normalize.cpp
rrawther: Also I think I mentioned this before. We shouldn't do data conversions here. vx_rpp has to be a thin layer which passes data to RPP without manipulating it

amd_openvx_extensions/amd_rpp/source/tensor/Transpose.cpp
SundarRajan28: Added comments for each assignment mentioning what value we are assigning to the ROI buffer

amd_openvx_extensions/amd_rpp/source/tensor/Normalize.cpp
```diff
@@ -55,19 +58,27 @@ static vx_status VX_CALLBACK refreshNormalize(vx_node node, const vx_reference *
         STATUS_ERROR_CHECK(vxQueryTensor((vx_tensor)parameters[1], VX_TENSOR_BUFFER_HIP, &roi_tensor_ptr, sizeof(roi_tensor_ptr)));
         STATUS_ERROR_CHECK(vxQueryTensor((vx_tensor)parameters[2], VX_TENSOR_BUFFER_HIP, &data->pDst, sizeof(data->pDst)));
         STATUS_ERROR_CHECK(vxQueryTensor((vx_tensor)parameters[3], VX_TENSOR_BUFFER_HIP, &roi_tensor_ptr_dst, sizeof(roi_tensor_ptr_dst)));
+        if ((numDims == 4) && (!data->pSrcDims)) {  // For NHWC/NCHW layouts, allocate pSrcDims pinned memory buffer to store ROI for all dims
```
rrawther, line 61: Why do we need to check if numDims == 4 here? The memory is allocated based on actual roiDims queried from tensor. Please remove the check

amd_openvx_extensions/amd_rpp/source/tensor/Normalize.cpp
```diff
@@ -55,19 +58,27 @@ static vx_status VX_CALLBACK refreshNormalize(vx_node node, const vx_reference *
...
+            if (err != hipSuccess)
+                return ERRMSG(VX_ERROR_NOT_ALLOCATED, "refresh: hipHostMalloc of size %ld failed \n", data->inputTensorDims[0] * roiDims * 2);
+        }
 #endif
     } else if (data->deviceType == AGO_TARGET_AFFINITY_CPU) {
         STATUS_ERROR_CHECK(vxQueryTensor((vx_tensor)parameters[0], VX_TENSOR_BUFFER_HOST, &data->pSrc, sizeof(data->pSrc)));
         STATUS_ERROR_CHECK(vxQueryTensor((vx_tensor)parameters[1], VX_TENSOR_BUFFER_HOST, &roi_tensor_ptr, sizeof(roi_tensor_ptr)));
         STATUS_ERROR_CHECK(vxQueryTensor((vx_tensor)parameters[2], VX_TENSOR_BUFFER_HOST, &data->pDst, sizeof(data->pDst)));
         STATUS_ERROR_CHECK(vxQueryTensor((vx_tensor)parameters[3], VX_TENSOR_BUFFER_HOST, &roi_tensor_ptr_dst, sizeof(roi_tensor_ptr_dst)));
+        if ((numDims == 3) && (data->inputTensorDims[2] == 1) && (!data->pSrcDims)) {  // For NHW/NTF/NFT layouts
```
rrawther, line 72: please remove check for numDims here

amd_openvx_extensions/amd_rpp/source/tensor/Normalize.cpp
```diff
@@ -55,19 +58,27 @@ static vx_status VX_CALLBACK refreshNormalize(vx_node node, const vx_reference *
...
-    RppSize_t numDims;
-    STATUS_ERROR_CHECK(vxQueryTensor((vx_tensor)parameters[0], VX_TENSOR_NUMBER_OF_DIMS, &numDims, sizeof(numDims)));
     STATUS_ERROR_CHECK(vxQueryTensor((vx_tensor)parameters[0], VX_TENSOR_DIMS, &data->inputTensorDims, sizeof(vx_size) * numDims));
 
     // For Identifying if the input Tensor is 1D (excluding the Nth dimension) [ even if 2nd dim is 1 - The tensor is considered 1D ]
-    if ((numDims == 3) && (data->inputTensorDims[2] == 1)) {
+    if ((numDims == 3) && (data->inputTensorDims[2] == 1)) {  // For NHW/NFT/NTF layouts
         RpptROI *src_roi = reinterpret_cast<RpptROI *>(roi_tensor_ptr);
         RpptROI *dst_roi = reinterpret_cast<RpptROI *>(roi_tensor_ptr_dst);
         for (unsigned i = 0, j = 0; i < data->inputTensorDims[0]; i++, j += 2) {
```
rrawther, line 84: Isn't data->pSrcDims[j + 1] is start_y position? Why is it Width

amd_openvx_extensions/amd_rpp/source/tensor/Normalize.cpp
rrawther: Another confusion here is the way RPP handles the x,y,z indeces are reversed. For e.g, NHWC means [0] = c, [1] = H, [2]=W etc. Please discuss with Abishek why this is interpreted reversed format.

amd_openvx_extensions/amd_rpp/source/tensor/Normalize.cpp
```diff
@@ -77,17 +88,48 @@ static vx_status VX_CALLBACK refreshNormalize(vx_node node, const vx_reference *
...
+                data->pSrcDims[index + 5] = src_roi[i].xywhROI.roiWidth;   // EndW
+            }
+            dst_roi[i].xywhROI = src_roi[i].xywhROI;
+        }
+        data->pSrcRoi = static_cast<unsigned *>(data->pSrcDims);
+    } else {  // For other layouts which already has generic ROI
         data->pSrcRoi = static_cast<unsigned *>(roi_tensor_ptr);
+        Rpp32u *dst_roi = static_cast<unsigned *>(roi_tensor_ptr_dst);
+        for (unsigned i = 0; i < data->inputTensorDims[0]; i++) {
+            unsigned index = i * roiDims * 2;
```
rrawther, line 119: What is *2 corresponds to?

amd_openvx_extensions/amd_rpp/source/tensor/Transpose.cpp
rrawther: Added few more comments in normalize which applies here too

amd_openvx_extensions/amd_rpp/source/tensor/Normalize.cpp
SundarRajan28: This check is added becoz for numDims = 3 (audio usecase - NHW/NFT/NTF) layout, the srcRoi buffer is created as BS * 2, not as BS * roiDims * 2. That's why there are different conditions added for numDims 3 and 4

amd_openvx_extensions/amd_rpp/source/tensor/Normalize.cpp
SundarRajan28: This check is added becoz for numDims = 3 (audio usecase - NHW/NFT/NTF) layout, the srcRoi buffer is created as BS * 2, not as BS * roiDims * 2. That's why there are different conditions added for numDims 3 and 4

amd_openvx_extensions/amd_rpp/source/tensor/Normalize.cpp
SundarRajan28: For NHWC/NCHW layouts, roiDims value is 3. We need to create a buffer for BS * 3 * 2 to hold start and end values for HWC values

amd_openvx_extensions/amd_rpp/source/tensor/Normalize.cpp
SundarRajan28: For audio layouts, roiDims is calculated differently. Removed the checks here and added the checks to assign roiDims accordingly

amd_openvx_extensions/amd_rpp/source/tensor/Normalize.cpp
SundarRajan28: Removed the checks here and added the checks to assign roiDims accordingly

amd_openvx_extensions/amd_rpp/source/tensor/Transpose.cpp
SundarRajan28: Added changes for both transpose and normalize according to the comments

amd_openvx_extensions/amd_rpp/source/tensor/Normalize.cpp
SundarRajan28: For audio tensors, if the 2nd dimension is 1 (single channel), the tensor is considered as 1D and the ROI is expected as such. So the ROI is only passed for the first dimension

rrawther (approved):

kiritigowda (approved):

PR #1497, opened 2025-03-21
CMakeLists - Set CXX STD - 17

kiritigowda:
* Set CXX default std to 17 from 14
* Set enforce std
* add cxx extensions


amd_openvx/openvx/CMakeLists.txt
```diff
@@ -159,7 +161,7 @@ if(WIN32)
     set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /MD /DVX_API_ENTRY=__declspec(dllexport)")
     set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /MDd /DVX_API_ENTRY=__declspec(dllexport)")
 else()
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -msse4.2 -std=gnu++14")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -msse4.2 -std=gnu++17")
```
scchan, line 164: Normally you shouldn't add the C++ flag here since CMAKE_CXX_STANDARD has been specified to ver 17.  If you need the GNU C++ extension you should enable CXX_EXTENSIONS. 

Same comment applies to other cmakefiles in this PR.

LakshmiKumar23 (approved):

rrawther (commented):
@kiritigowda  please check if we need to explicitly set CXX_FLAGS as commented

amd_openvx/openvx/CMakeLists.txt
kiritigowda: @scchan can this be achieved globally by setting `set(CMAKE_CXX_EXTENSIONS ON)`? Also, we want the min cmake version to remain at 3.10, noticed that the extension set is available in 3.20+, can you take a look at this? Thanks!

scchan: I'm not sure what you mean by "globally".  It applies to the scope of this cmakefile file just like other cmake variables.
If you enable `CXX_EXTENSIONS`, it may change the flag from `-std=c++17` to `std=gnu++17`.

rrawther (approved):

LakshmiKumar23 (approved):

PR #1514, opened 2025-06-06
Docs: Add environment variable reference page

adeljo-amd:
This is part of the environment variable initiative, where we aim to create a ROCm environment variable index page that contains the links to every component's environment variable page. I extracted the environment variables from the source code, but had to guess a lot of the values and purposes of the variables.

It would be great to get feedback on what the expected values are (of if there are default values for some), and whether the environment variables listed in the page are important for the users.


adeljo-amd:
@kiritigowda Sorry, I did not realise this was marked as draft when it was ready all this time - is there any other feedback concerning the environment variables listed here?

kiritigowda (approved):

PR #1694, opened 2026-06-20
Fix HIP CannyEdgeDetector failure on repeated vxProcessGraph

### LINKED ISSUE #1693: HIP CannyEdgeDetector fails on repeated vxProcessGraph: Canny stack GPU counter not reset
## Summary

The HIP backend's `CannyEdgeDetector` fails on the **second and subsequent executions of the same verified graph**. The failure is caused by `agoGpuHipDataOutputAtomicSync()` not resetting the GPU-side Canny stack atomic counter after reading it. The equivalent OpenCL path already resets the counter, but the HIP path was missing this step.

## Environment

- Host: `avinya` (AMD RYZEN AI MAX+ PRO 395 w/ Radeon 8060S / gfx1151, 40 CUs)
- ROCm/HIP: 7.13
- MIVisionX: current `develop` / `master`

## Reproduction

A minimal reproducer creates a FHD U8 image, a Canny node with `VX_THRESHOLD_TYPE_RANGE`, and calls `vxProcessGraph` in a loop:

```cpp
for (int iter = 0; iter < 20; iter++) {
    if (vxProcessGraph(g) != VX_SUCCESS) { /* fails on iter 2 */ }
}
```

Full reproducer: `repro_canny_hip.cpp` (attached conceptually; can be provided).

`openvx-mark` also reproduces it reliably with `--vision-parity` or any multi-iteration Canny benchmark.

## Observed error

```text
[VX LOG] status=-1: ERROR: hipMemcpyDtoH() for stacktop => 1
[VX LOG] status=-1: ERROR: agoWaitForNodesCompletion: single node wait failed
[VX LOG] status=-1: ERROR: agoWaitForNodesCompletion failed (-1:(null))
AMD OpenVX: HIP Error Reported -- invalid argument
[VX LOG] status=-1: ERROR: kernel com.amd.openvx.CannySobel_U16_U8_3x3_L1NORM exec failed (-2:(null))
```

`hipError_t == 1` is `hipErrorInvalidValue`.

## Root cause

`Hip_CannySuppThreshold_U8XY_U16_3x3` in `hipvx/vision_kernels.cpp` uses `atomicAdd((uint *)xyStack, stack_icount)` where `xyStack[0]` is the counter and coordinate data starts at `xyStackOffset` (= `DATA_GPU_ARRAY_OFFSET` = 16).

After the kernel finishes, `agoGpuHipDataOutputAtomicSync()` in `ago/ago_util_hip.cpp` reads the counter and copies the coordinates to host memory, but it **never writes 0 back to the GPU counter**. On the next graph execution the counter still holds the previous value, `atomicAdd` accumulates on top of it, and eventually `stackTop * sizeof(ago_coord2d_ushort_t)` exceeds the allocated buffer size, causing `hipMemcpyDtoH` to return `hipErrorInvalidValue`.

The OpenCL equivalent (`agoGpuOclDataOutputAtomicSync` in `ago_util_opencl.cpp`) correctly resets the counter:

```cpp
data->u.cannystack.stackTop = *(vx_uint32 *)stack;
*(vx_uint32 *)stack = 0;
```

## Proposed fix

In `amd_openvx/openvx/ago/ago_util_hip.cpp`, function `agoGpuHipDataOutputAtomicSync()`, after reading the coordinates and before returning, reset the GPU counter to 0:

```cpp
// reset the GPU atomic counter so the next graph execution starts from 0
if (data->hip_memory) {
    stime = agoGetClockCounter();
    vx_uint32 zero = 0;
    hipError_t err = hipMemcpyHtoD((void *)data->hip_memory, &zero, sizeof(vx_uint32));
    if (err) {
        agoAddLogEntry(&data->ref, VX_FAILURE, "ERROR: hipMemcpyHtoD() for stacktop reset => %d\n", err);
        return -1;
    }
    etime = agoGetClockCounter();
    graph->gpu_perf.buffer_write += etime - stime;
}
```

Additionally, clamp `stackTop` to `data->u.cannystack.count` before the coordinate read so an already-corrupted counter cannot read beyond the buffer.

## Verification of the fix

After applying the patch locally:

| Test | Result |
|:-----|:-------|
| Minimal Canny reproducer, 20 iterations FHD | all pass |
| `openvx-mark --vision-parity --resolution FHD` | 41/41 pass, 0 skipped, 0 failed |
| OpenVX CTS `vxuCanny.*:vxCanny.*` | 56/56 pass |
| OpenVX CTS `HarrisCorners.*` | 433/433 pass |
| Broad OpenVX CTS filter | 2366/2366 pass |

The secondary `HarrisCorners` error/slowness observed in the full benchmark was a downstream effect of the Canny failure corrupting the HIP stream; it disappears once Canny is fixed.

## Related observations

- `LaplacianPyramid` declares only `AGO_KERNEL_FLAG_DEVICE_CPU` on HIP/OpenCL, so it falls back to CPU inside a HIP graph and is slow. This is a separate missing-implementation issue.
- `HarrisCorners` is functional after the Canny fix, but its HIP performance on this APU for FHD is lower than CPU. This is a separate performance-tuning issue.

## Files involved

- `amd_openvx/openvx/ago/ago_util_hip.cpp` — location of the bug and fix
- `amd_openvx/openvx/ago/ago_util_opencl.cpp` — correct reference implementation of the counter reset
- `amd_openvx/openvx/hipvx/vision_kernels.cpp` — `Hip_CannySuppThreshold_U8XY_U16_3x3` kernel

I'd be happy to open a PR with the patch if the maintainers agree with the approach.

simonCatBot:
## Problem

On the HIP backend, calling `vxProcessGraph` more than once on a graph that contains a `CannyEdgeDetector` node eventually fails with errors such as `invalid argument` or `single node wait failed` from the Harris-Sobel/Canny stack kernels.

The root cause is that the HIP Canny stack output synchronization reads the GPU atomic counter in `agoGpuHipDataOutputAtomicSync()` but never resets it to zero. On the next graph execution the Canny kernel sees the previous `stackTop`, writes past the allocated stack buffer, and corrupts the following nodes.

The OpenCL backend already resets the counter after reading via map/unmap in `agoGpuOclDataOutputAtomicSync()`; the HIP path was missing the equivalent reset.

## Fix

In `amd_openvx/openvx/ago/ago_util_hip.cpp`:

1. Clamp `stackTop` to the Canny-stack capacity as a defensive guard.
2. After reading the stack contents, write zero back to the GPU atomic counter with `hipMemcpyHtoD()` so the next `vxProcessGraph` starts from a clean stack.

This matches the OpenCL implementation and fixes the repeated-execution failure.

Fixes: ROCm/MIVisionX#1693

## Testing

Built the HIP backend on an AMD RYZEN AI MAX+ PRO 395 / Radeon 8060S APU with:

```bash
cmake -DBACKEND=HIP -DGPU_SUPPORT=ON -DNEURAL_NET=OFF -DLOOM=OFF -DMIGRAPHX=OFF ..
make -j$(nproc)
```

Ran the full Khronos OpenVX 1.3 conformance suite (`OpenVX-cts`) from the `test_data` directory with the HIP libraries on `LD_LIBRARY_PATH`:

```
[ ALL DONE ] 5820 test(s) from 69 test case(s) ran
[ PASSED   ] 5820 test(s)
[ FAILED   ] 0 test(s)
[ DISABLED ] 8190 test(s)

Baseline:  863 required / 863 passed / 0 failed → PASSED
Vision:   4957 required / 4957 passed / 0 failed → PASSED
```

Also ran `openvx-mark --vision-parity` at FHD; the overall vision score is ~45,510 MP/s, on par with the pre-fix baseline (~45,782 MP/s), so there is no performance regression. `CannyEdgeDetector` runs at ~560 MP/s / ~3.7 ms at FHD.

## Impact

- Corrects Canny behavior on repeated graph execution for HIP.
- No change to OpenCL, CPU, or non-Canny code paths.
- No performance degradation observed.

### DIFF
--- a/amd_openvx/openvx/ago/ago_util_hip.cpp
+++ b/amd_openvx/openvx/ago/ago_util_hip.cpp
@@ -667,6 +667,10 @@ static int agoGpuHipDataOutputAtomicSync(AgoGraph * graph, AgoData * data) {
         }
         int64_t etime = agoGetClockCounter();
         graph->gpu_perf.buffer_read += etime - stime;
+        // clamp stackTop to capacity to prevent reading past the GPU buffer
+        if (stack > data->u.cannystack.count) {
+            stack = data->u.cannystack.count;
+        }
         data->u.cannystack.stackTop = stack;
         if (data->u.cannystack.stackTop > 0) {
             if (data->hip_memory) {
@@ -678,6 +682,19 @@ static int agoGpuHipDataOutputAtomicSync(AgoGraph * graph, AgoData * data) {
             }
             int64_t etime = agoGetClockCounter();
             graph->gpu_perf.buffer_read += etime - stime;
+        }
+        // reset the GPU counter to 0 so the next graph execution starts from a clean stack
+        // (matches the OpenCL path in agoGpuOclDataOutputAtomicSync)
+        if (data->hip_memory) {
+            vx_uint32 zero = 0;
+            stime = agoGetClockCounter();
+            hipError_t err = hipMemcpyHtoD(data->hip_memory, (void *)&zero, sizeof(vx_uint32));
+            if (err) {
+                agoAddLogEntry(&data->ref, VX_FAILURE, "ERROR: hipMemcpyHtoD() for stacktop reset => %d\n", err);
+                return -1;
+            }
+            etime = agoGetClockCounter();
+            graph->gpu_perf.buffer_write += etime - stime;
         }
     }
     return 0;


kiritigowda:
Fix captured in #1699

PR #1696, opened 2026-06-20
HIP: force HarrisCorners subgraph to CPU affinity to fix performance

simonCatBot:
## Problem

On the HIP backend the HarrisCorners subgraph was being scheduled on the GPU for the `HarrisSobel` and `HarrisScore` kernels, then falling back to the CPU-only `HarrisMergeSortAndPick` stage. This caused a full GPU→CPU readback of the F32 score image plus a slow CPU NMS/sort, making HIP HarrisCorners **~8× slower** than the CPU backend on the test APU (21 MP/s vs ~155 MP/s at FHD).

## Fix

Set `node->attr_affinity.device_type = AGO_KERNEL_FLAG_DEVICE_CPU` for the `HarrisCorners` subgraph on HIP. The existing `agoImportNodeConfig` in `agoDramaDivideAppend` propagates this affinity to the child nodes created during drama division (Sobel, Score, NMS, merge), so the **entire** Harris pipeline stays on the CPU. This matches the behavior of running with `AGO_DEFAULT_TARGET=CPU` and avoids the expensive GPU readback.

This change is HIP-only; OpenCL behavior is unchanged.

## Testing

Built the HIP backend on an AMD RYZEN AI MAX+ PRO 395 / Radeon 8060S APU with:

```bash
cmake -DBACKEND=HIP -DGPU_SUPPORT=ON -DNEURAL_NET=OFF -DLOOM=OFF -DMIGRAPHX=OFF ..
make -j$(nproc)
```

Ran the full Khronos OpenVX 1.3 conformance suite (`OpenVX-cts`) with the HIP libraries on `LD_LIBRARY_PATH`:

```
[ ALL DONE ] 5820 test(s) from 69 test case(s) ran
[ PASSED   ] 5820 test(s)
[ FAILED   ] 0 test(s)
[ DISABLED ] 8190 test(s)

Baseline:  863 required / 863 passed / 0 failed → PASSED
Vision:   4957 required / 4957 passed / 0 failed → PASSED
```

Ran `openvx-mark --vision-parity` at FHD:

| Metric | Before fix | After fix |
|:---|---:|---:|
| HarrisCorners MP/s | ~21.4 | **159.6** |
| HarrisCorners median (ms) | ~96.8 | **12.99** |
| Overall vision score | ~45,510 | **45,741** |

There is no overall performance regression; the other 40 vision kernels remain unchanged.

## Note

PR #1695 tried to expose GPU scheduling for HarrisCorners on HIP, but the GPU merge stage is not implemented, so the GPU path performed poorly. This PR takes the opposite, immediately beneficial approach: keep the Harris pipeline on the CPU for HIP until a full GPU merge/sort kernel exists. When that kernel lands, the affinity override can be removed.

### DIFF
--- a/amd_openvx/openvx/ago/ago_kernel_api.cpp
+++ b/amd_openvx/openvx/ago/ago_kernel_api.cpp
@@ -1891,6 +1891,16 @@ int ovxKernel_HarrisCorners(AgoNode * node, AgoKernelCommand cmd)
         node->metaList[6].data.u.arr.capacity = 0;
         node->metaList[7].data.u.scalar.type = VX_TYPE_SIZE;
         status = VX_SUCCESS;
+        // On HIP the HarrisCorners merge stage is CPU-only. Running the
+        // preceding Sobel/Score kernels on the GPU and then reading the score
+        // image back to the CPU for merge/sort is much slower than running the
+        // whole subgraph on the CPU on integrated APUs. Force CPU affinity for
+        // the HarrisCorners subgraph on HIP so all child nodes (Sobel, Score,
+        // NMS, merge) are scheduled on the CPU. Ago will propagate this
+        // affinity to the child nodes created during drama division.
+#if ENABLE_HIP
+        node->attr_affinity.device_type = AGO_KERNEL_FLAG_DEVICE_CPU;
+#endif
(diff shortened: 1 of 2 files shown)


kiritigowda:
Fix captured in #1699

PR #1731, opened 2026-08-05
Use tag to get linux runners

daniellegillai:
Changed some of the jobs to run on [controller-mode:docker] tag. others still using shark39.


kiritigowda (commented):
https://github.com/ROCm/MIVisionX/actions/runs/31049871486/job/92466106927#step:1:1 -- Linux-* may have worked

PR #1742, opened 2026-08-12
OpenVX: ROCTX/rocprof tracing enhancements

kiritigowda:
## Summary

Follow-up to #1739. Makes the ROCTX/rocprof tracing feature more meaningful for developers and closes gaps found in the initial implementation. All changes are opt-in behind `MIVISIONX_ENABLE_ROCPROF` (build) + `MIVISIONX_ROCPROF` (runtime); default builds are unaffected.

## Instrumentation gaps closed

- **GPU kernel time is now visible.** Per-node ranges only cover the asynchronous launch. Added a `MIVisionX: GPU sync/wait` range around the graph-completion `hipStreamSynchronize`/`clFinish`, where real device time lands.
- **Memory copies attributed to nodes.** Added `MIVisionX: copy-in <kernel> paramN` (H2D) on both the OpenCL and HIP input-sync paths, and `MIVisionX: copy-out <kernel>` (D2H) in both `agoDataSyncFromGpuToCpu` variants — so a developer can see which node/param drove a transfer instead of an anonymous copy event.
- **Pipelining/streaming instrumented.** Added `MIVisionX: pipelined execution` in `agoExecutePipelinedGraphOnce` (the common funnel for manual/auto/streaming), nesting the existing `agoExecuteGraph` range.

## Correctness / portability

- **Windows env var honored.** `enabled()` now reads `MIVISIONX_ROCPROF` via `agoGetEnvironmentVariable()` (GetEnvironmentVariableA on Windows) instead of `std::getenv`, which misses variables set through the Win32 API.
- **Thread-safe init.** Replaced the relaxed-atomic double-check in the runtime guard with a C++11 magic-static — once-only init with correct cross-thread ordering, less code.

## CI tooling

- **Memory tables always render.** The rocprof summary previously dropped the memory-copy/allocation tables silently when empty. They now show the data, an "N/A for the CPU backend" note, or a "no data captured — investigate" warning.
- **GPU jobs are gated.** The profiling workflow no longer runs GPU jobs on every PR. Post-merge pushes are path-filtered to tracing-relevant files; PRs run automatically on those paths or on demand via a `run-rocprof` label.

## Documentation

Added a "Profiling with rocprof / ROCTX" section to `amd_openvx/README.md`: build flag, runtime toggle, full `rocprofv3` command, a table explaining every range (with the async-launch caveat), Perfetto viewing, and the CI gating/label.

## Verification

Built and traced on both WSL (gfx1151) and a native ROCm box (santiago, gfx1100, RX 7900 XT):

- CPU-only build (macros compile away) — passes on both.
- HIP build with `-DMIVISIONX_ENABLE_ROCPROF=ON` — passes on both.
- On native GPU, `rocprofv3` confirmed the new ranges fire and correlate with the raw traces: `MIVisionX: GPU sync/wait` and `MIVisionX: copy-in ... param1/param2` appeared and matched the `MEMORY_COPY_HOST_TO_DEVICE` entries in the memory-copy trace, alongside existing per-node ranges, kernel dispatches, and allocations.

Notes: `copy-out` and `pipelined execution` compile and are placed identically to verified paths, but did not fire in the pure-GPU / non-pipelined test GDFs used — they need a mixed CPU/GPU-affinity graph and a pipelining graph respectively to exercise live.

## Test plan

- [ ] CI `ROCprof GDF profiling` workflow passes (gated run via `run-rocprof` label or tracing-path change)
- [ ] Default (non-ROCPROF) HIP and CPU builds unaffected

### DIFF
--- a/.github/scripts/run_rocprof_gdf.py
+++ b/.github/scripts/run_rocprof_gdf.py
@@ -531,9 +531,9 @@ def build_summary(
         )
         lines.append("")
 
+    lines.append("## Aggregate memory copies (host <-> device)")
+    lines.append("")
     if aggregate_copies:
-        lines.append("## Aggregate memory copies (host <-> device)")
-        lines.append("")
         lines.append("| Direction | Count | Total bytes | Total time |")
         lines.append("|-----------|------:|------------:|-----------:|")
         for direction, stat in sorted(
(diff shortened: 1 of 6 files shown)


kiritigowda:
image

AryanSalmanpour (approved):

